PR #38, opened 2019-08-11
RD-1051 - feat: add lambda-apigw-lambda support

saartochner:


src/parsers/aws.test.js
```diff
@@ -232,4 +232,19 @@ describe('aws parser', () => {
       },
     });
   });
+
+  test('awsParser -> happy flow', () => {
```
doriaviram-lumigo: This becomes to be the default parser, add test for without `x-amzn-requestid`

src/parsers/aws.js
```diff
@@ -62,3 +62,11 @@ export const kinesisParser = requestData => {
   const awsServiceData = { resourceName };
   return { awsServiceData };
 };
+
+export const awsParser = (requestData, responseData) => {
+  const { headers: resHeader } = responseData;
+  const messageId = resHeader ? (resHeader['x-amzn-requestid'] || {}) : undefined;
```
doriaviram-lumigo: Why {}? I think it should be undefined
its will clear on the tests :)

PR #49, opened 2019-09-22
feat: blacklist keys (security)

saartochner:


src/utils.js
```diff
@@ -313,3 +313,28 @@ export const parseQueryParams = queryParams => {
   });
   return obj;
 };
+
+const keyToOmitRegexes = () =>
+    JSON.parse(process.env.LUMIGO_BLACKLIST_REGEX || '[".*pass.*", ".*key.*"]').map(x => new RegExp(x, 'i'));
```
doriaviram-lumigo: Why not use includes? simpler and faster

src/utils.js
```diff
@@ -313,3 +313,28 @@ export const parseQueryParams = queryParams => {
...
+    } catch (e) {
+      return obj;
+    }
+  }
+  if (!obj || typeof obj !== "object") {
+    return obj;
+  }
+  return Object.keys(obj).reduce((newObj, key) => {
+    let value = omitKeys(obj[key]);
+    let shouldOmitKey = keyToOmitRegexes().some(regex => regex.test(key));
```
doriaviram-lumigo: You createing the regexes list every time. you can create this only once (if you decide to dont use simple contains )

PR #51, opened 2019-09-24
RD-1356 many to many

saartochner:


src/parsers/aws.js
```diff
@@ -54,12 +54,20 @@ export const sqsParser = requestData => {
...
-export const kinesisParser = requestData => {
+export const kinesisParser = (requestData, responseData) => {
   const { body: reqBody } = requestData;
+  const { body: resBody } = responseData;
   const reqBodyJSON = (!!reqBody && JSON.parse(reqBody)) || {};
+  const resBodyJSON = (!!resBody && JSON.parse(resBody)) || {};
   const resourceName =
     (reqBodyJSON['StreamName'] && reqBodyJSON.StreamName) || undefined;
   const awsServiceData = { resourceName };
+  if (resBodyJSON["SequenceNumber"]) {
```
doriaviram-lumigo: You can use here your own `getKinesisData` function?

doriaviram-lumigo: Okay, so at least the huge line out to a function

PR #52, opened 2019-09-29
fix: recognize if lambda is warm bug RD-1386

doriaviram-lumigo:


src/spans/awsSpan.js
```diff
@@ -67,7 +67,7 @@ export const getBasicSpan = () => {
   const account = awsAccountId;
 
   const readiness = isWarm() ? 'warm' : 'cold';
-  if (readiness === 'warm') {
+  if (readiness === 'cold') {
```
nirLumigo: Don't we need to `setCold` in this case?

src/spans/awsSpan.test.js
```diff
@@ -113,6 +113,12 @@ describe('awsSpan', () => {
     expect(awsSpan.getBasicSpan()).toEqual(expectedBasicSpan);
   });
 
+  test('getBasicSpan turn is warm', () => {
+    expect(utils.isWarm()).toEqual(false);
+    awsSpan.getBasicSpan();
```
nirLumigo: Why `getBasicSpan()` should change it to be `'warm'`?

doriaviram-lumigo: for the next time, you run the tracer, it was the behavior that Sagi wants, he just had a bug

src/spans/awsSpan.js
doriaviram-lumigo: if the lambda is cold, the next time you want its to be warm (not opposite)

PR #53, opened 2019-10-17
Add default blacklist

efimk-lu:


efimk-lu:
@doriaviram-lumigo why specifically through the ENV? There is a test that checks that the env is working + test that checks that these keys are valid

PR #55, opened 2019-10-27
fix: improve logger RD-1572

doriaviram-lumigo:


src/logger.js
```diff
@@ -1,9 +1,66 @@
 import { isDebug } from './utils';
 
 const LOG_PREFIX = '#LUMIGO#';
+const DUPLICATE_LOGS_TO_EMERGENCY_MODE = 3;
 
-export const invokeLog = type => (msg, obj = undefined) =>
-  isDebug() && exports.log(type, msg, obj);
+const printedLogSet = new Set([]);
```
nirLumigo:
```suggestion
const PRINTED_LOG_SET = new Set([]);
```

src/logger.js
```diff
@@ -1,9 +1,66 @@
...
+const pendingLogSet = new Set([]);
```
nirLumigo:
```suggestion
const PENDING_LOG_SET = new Set([]);
```

src/logger.js
```diff
@@ -1,9 +1,66 @@
...
+let duplicateLogsCount = 0;
```
nirLumigo:
```suggestion
let DUPLOCATE_LOGS_COUNT = 0;
```

src/logger.js
```diff
@@ -1,9 +1,66 @@
...
+  }
+};
+
+export const safePrint = (printedLogSet, type, message, obj) => {
+  const logObj = buildLogObject(type, message, obj);
+  if (!printedLogSet.has(JSON.stringify(logObj))) {
+    exports.log(type, message, obj);
+    printedLogSet.add(JSON.stringify(logObj));
+  } else {
+    duplicateLogsCount++;
```
nirLumigo: We count the total duplicates instead of the duplicates for each log.
If we have 3 different logs, each one is printed twice, we'll still get `DUPLICATE_LOGS_TO_EMERGENCY_MODE`.

src/logger.js
```diff
@@ -1,9 +1,66 @@
...
+};
+
+export const handlePendingLogs = (pendingLogSet, printedLogSet, logObj) => {
+  if (
+    !printedLogSet.has(JSON.stringify(logObj)) &&
+    !pendingLogSet.has(JSON.stringify(logObj))
+  ) {
+    pendingLogSet.add(JSON.stringify(logObj));
+  } else {
+    duplicateLogsCount++;
```
nirLumigo: The first `print` of each log isn't counted, means will get into `DUPLICATE_LOGS_TO_EMERGENCY_MODE` after 4 prints of the same log and not after 3. Is that what you meant?

src/logger.js
```diff
@@ -1,9 +1,66 @@
...
+export const isEmergencyMode = () =>
+  duplicateLogsCount >= DUPLICATE_LOGS_TO_EMERGENCY_MODE;
+
+export const resetLogger = () => {
+  duplicateLogsCount = 0;
+  printedLogSet.clear();
+  pendingLogSet.clear();
+};
+
+export const handlePendingLogs = (pendingLogSet, printedLogSet, logObj) => {
```
nirLumigo:
```suggestion
export const addToPendingLogs = (pendingLogSet, printedLogSet, logObj) => {
```

src/logger.js
```diff
@@ -1,9 +1,66 @@
 import { isDebug } from './utils';
 
 const LOG_PREFIX = '#LUMIGO#';
+const DUPLICATE_LOGS_TO_EMERGENCY_MODE = 3;
```
nirLumigo: A higher number (like 10) might be safer and get the same results, isn't it?

PR #57, opened 2019-10-28
fix: circular object RD-1572

doriaviram-lumigo:


src/tools/noCirculars.js
```diff
@@ -0,0 +1,17 @@
+const noCirculars = v => {
```
doriaviram-lumigo: Copied (And add some nice tests)

src/tools/noCirculars.test.js
```diff
@@ -0,0 +1,72 @@
...
+      a.b = b;
+      let c = noCirculars(a);
+      expect(c).toEqual({
+        b: {
+          a: '[Circular]',
+        },
+      });
+    });
+
+    it('an inherited property', function() {
```
saartochner: delete

src/tools/noCirculars.test.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+import { noCirculars } from './noCirculars';
+
+describe('noCirculars', function() {
+  beforeEach(function() {
```
saartochner: delete

src/tools/noCirculars.test.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+      Child.prototype = new Base();
+
+      let z = noCirculars(new Child());
+      expect(z).toHaveProperty('child', true);
+      expect(z).not.toHaveProperty('base');
+    });
+
+    it('an array-like object', function() {
+      let t = { length: 3, 0: 'test', 1: 'test', 2: 'test' };
```
saartochner:
```suggestion
      let t = ['test', 'test2', 'test3'];
```
and check a circular example

PR #58, opened 2019-10-29
RD-1580 - feat: warn the user on execution leaks and don't crush the tracer

saartochner:


src/utils.js
```diff
@@ -343,3 +344,13 @@ export const omitKeys = obj => {
     return newObj;
   }, {});
 };
+
+export const lumigoWarnings = msg => {
```
doriaviram-lumigo: move to logger

doriaviram-lumigo:
```suggestion
export const warnClient = msg => {
```

src/utils.js
```diff
@@ -343,3 +344,13 @@ export const omitKeys = obj => {
...
+  if (process.env.LUMIGO_WARNINGS === "off") {
+    logger.debug('Does not warn the user about', msg);
+    return false;
+  }
+  // eslint-disable-next-line no-console
+  console.log("Lumigo Warning:", msg);
```
doriaviram-lumigo: export format to const

src/tracer.js
```diff
@@ -85,7 +87,13 @@ export const sendEndTraceSpans = async (functionSpan, handlerReturnValue) => {
   const spans = SpansContainer.getSpans();
   await sendSpans(spans);
   logger.debug('Tracer ended');
-  clearGlobals();
+  const currentTransactionId = getCurrentTransactionId();
+  if (spans.some(s => s.transactionId !== currentTransactionId)){
+    lumigoWarnings("Code leak detected. More information is available in: ");
```
doriaviram-lumigo: add doc

PR #59, opened 2019-10-31
fix: invalid return value in kinesis

efimk-lu:


src/parsers/aws.js
```diff
@@ -1,6 +1,6 @@
 import { parseQueryParams } from '../utils';
 import parseXml from '../tools/xmlToJson';
-
+import * as logger from '../logger';
```
saartochner: I'll be happy to see a line break after the imports

PR #60, opened 2019-11-03
feat: change the event size

efimk-lu:


src/spans/awsSpan.js
```diff
@@ -11,7 +11,7 @@ import {
   stringifyAndPrune,
   isAwsService,
   parseErrorObject,
-  omitKeys,
+  omitKeys, getEventEntitySize,
```
saartochner: did you pretiffy?

src/utils.js
```diff
@@ -173,6 +173,14 @@ export const isString = x =>
 
 export const MAX_ENTITY_SIZE = 1024;
 
+export const getEventEntitySize = () => {
+  if (!isNaN(parseInt(process.env['MAX_EVENT_ENTITY_SIZE']))) {
```
saartochner: '''
return parseInt(process.env['MAX_EVENT_ENTITY_SIZE']) || MAX_ENTITY_SIZE
'''

src/utils.test.js
```diff
@@ -190,6 +191,21 @@ describe('utils', () => {
...
+    process.env = { ...oldEnv };
+  });
+
+  test('getEventEntitySize NaN', () => {
+    const oldEnv = Object.assign({}, process.env);
+    process.env = { ...oldEnv, MAX_EVENT_ENTITY_SIZE: 'A 2048' };
+    expect(utils.getEventEntitySize()).toBe(MAX_ENTITY_SIZE);
+    process.env = { ...oldEnv };
+  });
+
```
saartochner: can you also check the event+prune itself?
I.e.  give it a long event and check that it is pruned..

src/spans/awsSpan.js
efimk-lu: Mmm. It passed the tests. There was a bug in checks.sh and the circleci script. Fixing

src/utils.test.js
efimk-lu: Not following you, what's wrong with the prune tests?

saartochner: you only checked the function `getEventEntitySize`, and not the actual change in `getFunctionSpan`

PR #63, opened 2019-11-12
RD-1514 - feat: wrap only once + autoinstrument handler + token and edge from envs

saartochner:


efimk-lu:
@saartochner please do not merge this one yet

.circleci/config.yml
```diff
@@ -54,6 +54,9 @@ jobs:
       - run:
           name: install-npm-wee
           command: npm install
+      - run:
+          name: install-npm-inner
+          command: cd auto_instrument_handler; npm install
```
doriaviram: I think its better to use pushd and popd, after this the circle while be on `auto_instrument_handler` folder

src/globals.js
```diff
@@ -31,9 +31,9 @@ export const TracerGlobals = (() => {
     Object.assign(handlerInputs, { event: {}, context: {} });
 
   const setTracerInputs = ({
-    token = '',
+    token = process.env.LUMIGO_TRACER_TOKEN || '',
     debug = false,
```
doriaviram: Nice, change the debug mechanism to work also in this way.
This is a better way but its weird that we have 2 ways to handle env vars inputs

src/globals.js
```diff
@@ -31,9 +31,9 @@ export const TracerGlobals = (() => {
...
-    edgeHost = '',
+    edgeHost = process.env.LUMIGO_TRACER_HOST || '',
     switchOff = false,
```
doriaviram: Same I said to debug also to switchOff

src/tracer.js
```diff
@@ -133,6 +133,11 @@ export const trace = ({
     eventFilter,
   });
 
+  if (context.wrapped_by_lumigo) {
```
doriaviram: I think you will have a bug here, try to test this on a non-async function
and don't call the callback immediately. 

Just something to let you think of, the user create a non-async func and you return him an async func

src/tracer.js
```diff
@@ -133,6 +133,11 @@ export const trace = ({
...
+    return userHandler(event, context, callback);
+  }
+  context.wrapped_by_lumigo = true;
```
doriaviram: the user may see this, maybe we should start at `__` to make clear it's our private

src/tracer.test.js
```diff
@@ -392,4 +392,22 @@ describe('tracer', () => {
     );
     expect(spies.warnClient).toHaveBeenCalled();
   });
+
```
doriaviram: Add test for non async func

auto_instrument_handler/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "test": "jest"
+  },
+  "files": [
+    "index.js"
+  ],
+  "author": "saart",
+  "license": "ISC",
+  "devDependencies": {
+    "jest": "^24.9.0",
+    "@lumigo/tracer": "file:.."
```
doriaviram: This may be in dependencies and not dev (the user should upload this file to his func)

auto_instrument_handler/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "lumigo_auto_instrument",
```
doriaviram: node packages is with `package-name`

src/tracer.js
doriaviram: use camelCase

auto_instrument_handler/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+const handler = (event, context, callback) => {
+  const originalHandler = process.env[ORIGINAL_HANDLER_KEY];
+  if (!originalHandler) {
+    throw Error(
+      'Could not load the original handler. Are you sure that the handler is correct?'
+    );
+  }
+  const handlerParts = originalHandler.split('.');
+  const moduleName = `/var/task/${handlerParts.slice(0, -1).join('.')}`;
```
doriaviram: Maybe add a comment here of how the handler should look like

auto_instrument_handler/index.js
```diff
@@ -0,0 +1,26 @@
+const lumigo = require('@lumigo/tracer')({});
+
+const ORIGINAL_HANDLER_KEY = 'LUMIGO_ORIGINAL_HANDLER';
+
+const handler = (event, context, callback) => {
+  const originalHandler = process.env[ORIGINAL_HANDLER_KEY];
+  if (!originalHandler) {
```
doriaviram: Also, check the format and throw a readable error to the user, that he does not get an error on the split and slice functions

auto_instrument_handler/index.js
```diff
@@ -0,0 +1,26 @@
...
+    );
+  }
+  const handlerParts = originalHandler.split('.');
+  const moduleName = `/var/task/${handlerParts.slice(0, -1).join('.')}`;
+  const functionName = handlerParts.slice(-1);
+  let module;
+  try {
+    module = require(moduleName);
+  } catch (e) {
+    throw Error(
```
doriaviram: log in debug the moduleName you tried to find (i know you don't have the logger here) maybe just check the env var and print, your call

auto_instrument_handler/package.json
saartochner: I don't think so.
He shouldn't upload this package alone - he should just use the tracer, and it's weird that this package will be inside the node_modules of the package *and* in the node_modules of the layer...

src/tracer.js
saartochner: 😳

auto_instrument_handler/index.js
saartochner: Great idea!

auto_instrument_handler/index.js
saartochner: I'll put it inside the error message 👍

src/tracer.js
saartochner: I'm not sure what you meant,
I added a test and it worked.

Can we talk about it? :)

src/tracer.test.js
```diff
@@ -392,4 +392,32 @@ describe('tracer', () => {
...
+    let callBackCalled = false;
+    const callback = (err, val) => {
+      expect(val).toEqual(2);
+      callBackCalled = true;
+    };
+    const userHandlerSync = (event, context, callback) => {
+      setTimeout(() => {
+        callback(null, 2);
+      }, 0);
```
doriaviram: return here some dummy value and check he ignore that

PR #68, opened 2019-11-18
fix: improve http requests RD-1735

doriaviram-lumigo:


src/tracer.js
```diff
@@ -107,8 +84,6 @@ export const isCallbacked = handlerReturnValue => {
 export const endTrace = async (functionSpan, handlerReturnValue) => {
   try {
     if (functionSpan && !isSwitchedOff() && isAwsEnvironment()) {
-      // eslint-disable-next-line no-undef
-      TracerGlobals.timeoutTimer && clearTimeout(TracerGlobals.timeoutTimer);
```
efimk-lu: Shouldn't you remove also `clearTimeout`?

src/tracer.test.js
```diff
@@ -39,6 +37,8 @@ describe('tracer', () => {
   spies.clearGlobals = jest.spyOn(globals, 'clearGlobals');
   spies.warnClient = jest.spyOn(logger, 'warnClient');
   spies.logFatal = jest.spyOn(logger, 'fatal');
+  spies.log = jest.spyOn(console, 'log');
```
efimk-lu: Why did you add these spies?

src/tracer.js
doriaviram-lumigo: There is not TracerGlobals.timeoutTimer any more

src/tracer.test.js
doriaviram-lumigo: Now he dont print logs when you run npm run test

src/tracer.js
```diff
@@ -61,42 +58,18 @@ export const startTrace = async () => {
...
-    SpansContainer.clearSpans();
-  } else {
-    clearGlobals();
-  }
+  const spansToSend = spans.filter(
+    span => span.transactionId === currentTransactionId
+  );
+  await sendSpans(spansToSend);
+  const hasSpansFromPreviousInvocation = spansToSend.length !== spans.length;
+  hasSpansFromPreviousInvocation && logger.warnClient(LEAK_MESSAGE);
```
uri-p: Add debug print of the spans from the previous invocation

PR #71, opened 2019-12-10
fix: exceptions handling

nirLumigo:


src/logger.js
```diff
@@ -1,10 +1,19 @@
-import { isDebug } from './utils';
+import { TracerGlobals } from './globals';
 
 const LOG_PREFIX = '#LUMIGO#';
 const WARN_CLIENT_PREFIX = 'Lumigo Warning';
 
+export const isDebug = () => {
```
nirLumigo: Moved this function from utils.js because now utils.js uses logger.js and I wanted to avoid circular import.
Also, I think it makes sense to get the log level from the logger.

src/logger.test.js
```diff
@@ -98,4 +98,12 @@ describe('logger', () => {
     process.env.LUMIGO_WARNINGS = undefined;
     expect(logger.warnClient('msg')).toEqual(true);
   });
+
+  test('isDebug', () => {
```
nirLumigo: Moved from utils.test.js

src/hooks/http.js
```diff
@@ -78,25 +79,32 @@ export const wrappedHttpResponseCallback = (
...
-      receivedTime,
-      body,
-      headers: lowerCaseObjectKeys(headers),
-    };
-    const fixedRequestData = noCirculars(requestData);
-    const fixedResponseData = noCirculars(responseData);
-    const httpSpan = getHttpSpan(fixedRequestData, fixedResponseData);
-    SpansContainer.addSpan(httpSpan);
-  });
+  try {
```
doriaviram: Why you don't use here safeExcute ?

src/hooks/http.js
```diff
@@ -110,12 +118,17 @@ export const httpRequestEndWrapper = requestData => originalEndFn =>
 export const httpRequestOnWrapper = requestData => originalOnFn =>
   function(event, callback) {
     if (event === 'response' && callback && !callback.__lumigoSentinel) {
-      const wrappedCallback = exports.wrappedHttpResponseCallback(
-        requestData,
-        callback
-      );
-      wrappedCallback.__lumigoSentinel = true;
-
+      let wrappedCallback;
```
doriaviram:
```suggestion
      let wrappedCallback = callback;
```

src/hooks/http.js
```diff
@@ -110,12 +118,17 @@ export const httpRequestEndWrapper = requestData => originalEndFn =>
...
+        wrappedCallback = exports.wrappedHttpResponseCallback(
+          requestData,
+          callback
+        );
+        wrappedCallback.__lumigoSentinel = true;
+      } catch (err) {
+        logger.warn('Failed at wrapping with wrappedHttpResponseCallback', err);
+        return originalOnFn.apply(this, [event, callback]);
+      }
       return originalOnFn.apply(this, [event, wrappedCallback]);
```
doriaviram: Make return only once

src/hooks/http.js
```diff
@@ -184,10 +197,17 @@ export const isAlreadyTraced = callback =>
...
-    const { url, options, callback } = httpRequestArguments(args);
-    const host = getHostFromOptionsOrUrl(options, url);
+    let isTraceDisabled, url, options, callback, host;
+    try {
+      ({ url, options, callback } = httpRequestArguments(args));
+      host = getHostFromOptionsOrUrl(options, url);
+      isTraceDisabled = isBlacklisted(host) || isAlreadyTraced(callback);
+    } catch (err) {
+      logger.warn('request parsing error', err);
+      isTraceDisabled = true;
```
doriaviram: Maybe its better the tracer "starting" disabled and you try to change it. if you failed its stay disabled

src/logger.js
doriaviram: 🚀

src/logger.test.js
doriaviram: 🎸

src/utils.test.js
```diff
@@ -664,4 +668,15 @@ describe('utils', () => {
     const nullObject = null;
     expect(omitKeys(nullObject)).toEqual(null);
   });
+
+  test('safeExecute run function', () => {
+    expect(safeExecute(() => 5)()).toEqual(5);
+  });
+
+  test('safeExecute catch exception', () => {
```
doriaviram: Add test for async func

src/tracer.js
```diff
@@ -124,16 +124,20 @@ export const trace = ({
...
-  TracerGlobals.setTracerInputs({
-    token,
-    debug,
-    edgeHost,
-    switchOff,
-    eventFilter,
-  });
-
-  startHooks();
+  try {
```
doriaviram: safeExcute?

src/utils.js
```diff
@@ -351,3 +348,14 @@ export const omitKeys = obj => {
     return newObj;
   }, {});
 };
+
+export const safeExecute = (
+  callback,
+  message = 'Error in Lumigo tracer'
+) => () => {
```
doriaviram: Why you return a func?

src/utils.js
```diff
@@ -351,3 +348,14 @@ export const omitKeys = obj => {
     return newObj;
   }, {});
 };
+
+export const safeExecute = (
```
doriaviram: support async to

src/hooks/http.js
nirLumigo: `safeExecute` works only on functions, I can't use it on specific like context-manager.
I don't want to use `safeExecute` on the entire function because I don't want to catch user's errors.

src/utils.js
nirLumigo: If we won't return a function, `safeExecute(func)` will run `func()`.
But in some cases, we don't want to run `func()` now but to pass it as a callback.

src/tracer.js
nirLumigo: This is the "main", here we also run the code of the user and we don't want to catch his exceptions.

src/utils.js
nirLumigo: There is still no async function that needs `safeExecute`. We can add async support if we'll need it in the future :)

src/utils.test.js
nirLumigo: Look at the comment of adding async support :)

src/tracer.js
doriaviram: 👍

src/utils.js
doriaviram: 👍

src/utils.test.js
doriaviram: 🎸

PR #74, opened 2019-12-29
Rd 1972

efimk-lu:


scripts/bd_to_prod.sh
```diff
@@ -32,15 +32,15 @@ echo "Creating layer file"
 ./scripts/prepare_layer_files.sh
 
 echo "Creating lumigo-node layer"
-../utils/common_bash/create_layer.sh lumigo-node-tracer ALL nodejs "nodejs10.x nodejs8.10"
+../utils/common_bash/create_layer.sh lumigo-node-tracer ALL nodejs "nodejs10.x node12.x"
```
saartochner:
```suggestion
../utils/common_bash/create_layer.sh lumigo-node-tracer ALL nodejs "nodejs10.x nodejs12.x"
```?

Comment (nirLumigo) [inline scripts/bd_to_prod.sh]:
```suggestion
../utils/common_bash/create_layer.sh lumigo-node-tracer ALL nodejs "nodejs10.x nodejs12.x"
```
It was with `js` before and it is also with `js` in other places.

efimk-lu:
@saartochner / @nirLumigo created the 12.X layer locally and tested it on a Lambda.
@saartochner does it affect auto instrumentation?

saartochner:
> @saartochner / @nirLumigo created the 12.X layer locally and tested it on a Lambda.
> @saartochner does it affect auto instrumentation?

nope

PR #76, opened 2020-01-02
feat: release with delay

nirLumigo:


.circleci/config.yml
```diff
@@ -99,6 +103,26 @@ jobs:
       - run:
           name: run tests
           command: cd ../integration-tests && export testNode=true && npm run test
+  deploy-beta:
```
nirLumigo: Same as `deploy` except the last line

scripts/bd_to_beta.sh
```diff
@@ -0,0 +1,29 @@
...
+echo "Pushing beta version to npm"
+echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > .npmrc
+BETA_VERSION="$(npm view @lumigo/tracer dist-tags.beta)"
+BETA_VERSION=${BETA_VERSION:-0.0.0}
+npm version --no-git-tag-version $BETA_VERSION
+npm version --no-git-tag-version prepatch --preid beta
+npm publish --tag beta
+rm .npmrc
+
+echo "Release with delay"
```
nirLumigo: All the other parts were copied from bd_to_prod.sh

.circleci/config.yml
efimk-lu: Both deploy to beta and regular deploy will run when merging to master, that's not what you want to happen, right?

efimk-lu: The `deploy` should run only on trigger

nirLumigo: Yes, but the flow for triggering later deploy is still not done.

PR #81, opened 2020-01-06
Rd 1989 send only specific aliases

GuyMoses:


efimk-lu:
@GuyMoses in addition to Dori please ask @nirLumigo to do CR for you as well

src/utils.js
```diff
@@ -136,7 +136,30 @@ export const isPruneTraceOff = () =>
   );
 
 export const isSwitchedOff = () => {
-  return TracerGlobals.getTracerInputs().switchOff;
+  return TracerGlobals.getTracerInputs().switchOff || !isValidAlias();
+};
+
+export const getValidAliasesOrEmptyArray = () =>
+  process.env['VALID_ALIASES'] || [];
```
doriaviram: Talk with Efi about the name, maybe should be
```suggestion
  process.env['LUMIGO_VALID_ALIASES'] || [];
```

src/spans/awsSpan.js
```diff
@@ -60,6 +62,8 @@ export const getBasicSpan = () => {
   const transactionId = getCurrentTransactionId();
 
   const awsAccountId = getAccountId(lambdaContext);
+  const invokedArn = getInvokedArn();
```
efimk-lu: What happens if there is an exception thrown from the method, what will happen then? Is it safeguard?

src/utils.js
```diff
@@ -136,7 +136,30 @@ export const isPruneTraceOff = () =>
...
+  getHandlerContext().functionVersion || '';
+
+export const getInvokedAliasOrNull = () =>
+  safeExecute(() => {
+    return getInvokedArn().split(':').length >= 8
+      ? getInvokedArn().split(':')[7]
+      : null;
+  })();
+
+export const isValidAlias = () => {
```
efimk-lu: Missing logs that will help us to debug

src/utils.js
```diff
@@ -151,6 +174,9 @@ export const setVerboseMode = () => (process.env['LUMIGO_VERBOSE'] = 'TRUE');
 
 export const setSwitchOff = () => (process.env['LUMIGO_SWITCH_OFF'] = 'TRUE');
 
+export const setValidAliases = aliases =>
```
efimk-lu: I can see it's used just for testing, can't you change the value of the environment variable before the test?

src/utils.js
```diff
@@ -136,7 +136,30 @@ export const isPruneTraceOff = () =>
...
+  safeExecute(() => {
+    return getInvokedArn().split(':').length >= 8
+      ? getInvokedArn().split(':')[7]
+      : null;
+  })();
+
+export const isValidAlias = () => {
+  const validAliases = getValidAliasesOrEmptyArray();
+  const currentAlias = getInvokedAliasOrNull();
+  return validAliases.length === 0 || validAliases.includes(currentAlias);
```
efimk-lu: The aliases here are `contain`, so if the user has two aliases say `prod` and `noProd` if they will use `prod` they will catch both aliases, right?
I don't mind having it, but please make sure the customer does not have aliases with the structure I mentioned

src/utils.js
```diff
@@ -136,7 +136,30 @@ export const isPruneTraceOff = () =>
   );
 
 export const isSwitchedOff = () => {
-  return TracerGlobals.getTracerInputs().switchOff;
+  return TracerGlobals.getTracerInputs().switchOff || !isValidAlias();
+};
+
+export const getValidAliasesOrEmptyArray = () =>
```
efimk-lu: The function returns two types:
* Array
* A string

It's confusing, return a single type

src/utils.js
```diff
@@ -136,7 +136,30 @@ export const isPruneTraceOff = () =>
...
+
+export const getHandlerContext = () =>
+  TracerGlobals.getHandlerInputs().context || {};
+
+export const getInvokedArn = () => getHandlerContext().invokedFunctionArn || '';
+export const getInvokedVersion = () =>
+  getHandlerContext().functionVersion || '';
+
+export const getInvokedAliasOrNull = () =>
+  safeExecute(() => {
```
efimk-lu: Do you have a test that checks for invalid ARN?

src/utils.js
nirLumigo: Why an empty array? If `process.env['VALID_ALIASES']` exists then you return a `string`, isn't it? I think that ENV variables are usually strings.

If anyway an array is returned then we can rename it:
```suggestion
export const getValidAliases = () =>
```

src/spans/awsSpan.js
GuyMoses: It's running inside `safeExecute`

src/utils.js
nirLumigo: This means that when there isn't `validAliases`, the value will be `true`. Is this the behavior we want?

GuyMoses: I don't understand

src/utils.test.js
```diff
@@ -229,6 +229,95 @@ describe('utils', () => {
...
+        invokedFunctionArn: 'arn:aws:lambda:region:account:function:name:alias',
+      },
+    });
+    utils.setValidAliases(['wrong']);
+    expect(utils.isSwitchedOff()).toBe(true);
+    TracerGlobals.clearHandlerInputs();
+    process.env = { ...oldEnv };
+  });
+
+  test('isSwitchedOffValidAlias', () => {
```
nirLumigo: Same test twice

src/utils.js
GuyMoses: no, that is why it runs inside `safeExecute`

src/utils.test.js
```diff
@@ -229,6 +229,95 @@ describe('utils', () => {
...
+      event: {},
+      context: {
+        invokedFunctionArn: 'arn:aws:lambda:region:account:function:name:3',
+      },
+    });
+    utils.setValidAliases([]);
+    expect(utils.isValidAlias()).toEqual(true);
+    utils.setValidAliases(['1', '2']);
+    expect(utils.isValidAlias()).toEqual(false);
+    utils.setValidAliases(['1', '2', '3']);
```
nirLumigo: Put `3` in a variable so it will be clear when it should be `false` and `true`. Use it also in `invokedFunctionArn`.
```suggestion
    utils.setValidAliases(['1', '2', currentAlias]);
```

src/utils.js
efimk-lu: But add a test

src/utils.js
efimk-lu: It's related to `getValidAliasesOrEmptyArray` and on the return value, if you are returning string then includes is like contains `XYZ.contains(XY) is true`, but `['XYZ'].contains('XY') is false`

src/utils.test.js
```diff
@@ -238,7 +238,7 @@ describe('utils', () => {
         invokedFunctionArn: 'arn:aws:lambda:region:account:function:name:alias',
       },
     });
-    utils.setValidAliases(['wrong']);
+    process.env['LUMIGO_VALID_ALIASES'] = '["wrong"]';
```
efimk-lu: It's a complicated way to write it, why not `wrong,right` and then split & trim it?

src/utils.js
```diff
@@ -136,7 +136,35 @@ export const isPruneTraceOff = () =>
   );
 
 export const isSwitchedOff = () => {
-  return TracerGlobals.getTracerInputs().switchOff;
+  return TracerGlobals.getTracerInputs().switchOff || !isValidAlias();
```
efimk-lu: I'm missing logs telling that aliases are in use

GuyMoses: if anything will be filtered by the aliases - there's an info log inside the filter function

src/utils.test.js
GuyMoses: I don't understand

src/utils.js
GuyMoses: fixed that, only arrays now

src/utils.test.js
GuyMoses: in our documentation, it's the way I wrote it

src/utils.js
GuyMoses: added

PR #85, opened 2020-01-27
docs: update layer version

efimk-lu:


scripts/bd_to_prod.sh
```diff
@@ -31,6 +31,10 @@ echo ${KEY} | gpg --batch -d --passphrase-fd 0 ${enc_location} > ~/.aws/credenti
 echo "Creating layer file"
 ./scripts/prepare_layer_files.sh
 
+echo "Push to NPM"
```
saartochner: Here you ill upload the package with the wrong layer arn (in layers/LAYERS10x.md for example).
It may damage you if we will send the layer arn through parsing this file (but maybe we will never do it because of the tracer's version) --> you to decide...

PR #93, opened 2020-03-22
feat: add step function capabilities to the tracer

saartochner:
Two simple additions:
* when the step finishes, we add a new key to the event that contains a special messageId
* when the step begins, we extract from the event the messageId (search for it recursively, because the step function configuration can change its position)

### DIFF
--- a/src/events.test.js
+++ b/src/events.test.js
@@ -1,3 +1,5 @@
+import * as utils from './utils';
+
 const events = require('./events');
 const exampleS3Event = require('./testdata/events/s3-event.json');
 const exampleSnsEvent = require('./testdata/events/sns-event.json');
@@ -106,6 +108,18 @@ describe('events', () => {
     expect(events.getEventInfo(exampleS3Event)).toEqual({
       arn: 'arn:aws:s3:::mybucket',
       triggeredBy: 's3',
+    });
+
+    jest.spyOn(utils, 'isStepFunction');
+    utils.isStepFunction.mockReturnValueOnce(true);
+    expect(
+      events.getEventInfo({
+        data: 1,
+        [utils.LUMIGO_EVENT_KEY]: { [utils.STEP_FUNCTION_UID_KEY]: '123' },
+      })
+    ).toEqual({
+      triggeredBy: 'stepFunction',
+      messageId: '123',
     });
   });
 });
--- a/src/events.test.js
+++ b/src/events.test.js
@@ -1,4 +1,5 @@
 import * as utils from './utils';
+import { TracerGlobals } from './globals';
 
 const events = require('./events');
 const exampleS3Event = require('./testdata/events/s3-event.json');
@@ -110,8 +111,7 @@ describe('events', () => {
       triggeredBy: 's3',
     });
 
-    jest.spyOn(utils, 'isStepFunction');
-    utils.isStepFunction.mockReturnValueOnce(true);
+    TracerGlobals.setTracerInputs({ stepFunction: true });
     expect(
       events.getEventInfo({
         data: 1,
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -5,7 +5,11 @@ import {
   getEdgeUrl,
   removeLumigoFromStacktrace,
   isSendOnlyIfErrors,
+  isStepFunction,
   safeExecute,
+  getRandomId,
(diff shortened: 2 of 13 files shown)


src/events.test.js
```diff
@@ -107,5 +109,17 @@ describe('events', () => {
       arn: 'arn:aws:s3:::mybucket',
       triggeredBy: 's3',
     });
+
+    jest.spyOn(utils, 'isStepFunction');
```
doriaviram: Don't mock this function, mock low-level as you can, override ENV var or something

src/utils.js
```diff
@@ -389,3 +394,23 @@ export const safeExecute = (
...
+      foundValue = noCircularEvent[k];
+      return true;
+    }
+    if (noCircularEvent[k] && typeof noCircularEvent[k] === 'object') {
+      foundValue = recursiveGetLumigoEventKey(noCircularEvent[k]);
+      return foundValue !== undefined;
+    }
+  });
+  return foundValue;
+};
```
doriaviram: I think we should create a custom function, and not running the object n*2. the tracer is slow enough

src/utils.js
```diff
@@ -389,3 +394,23 @@ export const safeExecute = (
...
 };
+
+export const getLumigoEventKey = event => {
+  const noCircularEvent = noCirculars(event);
+  return recursiveGetLumigoEventKey(noCircularEvent);
+};
+
+const recursiveGetLumigoEventKey = noCircularEvent => {
+  let foundValue = undefined;
+  Object.keys(noCircularEvent).some(function(k) {
```
doriaviram: Use arrow func (also you can save this to as an own function, right now you create new func for every key)

src/utils.js
```diff
@@ -389,3 +394,23 @@ export const safeExecute = (
     logger.warn(message, err);
   }
 };
+
+export const getLumigoEventKey = event => {
+  const noCircularEvent = noCirculars(event);
+  return recursiveGetLumigoEventKey(noCircularEvent);
+};
+
+const recursiveGetLumigoEventKey = noCircularEvent => {
```
doriaviram: Extract here to other function that recursive get key from an object, and get `LUMIGO_EVENT_KEY` as param, this code we cant resue :)

src/tracer.test.js
```diff
@@ -445,4 +451,52 @@ describe('tracer', () => {
     // No exception.
     expect(handler).toHaveBeenCalledOnce();
   });
+
+  test('performStepFunctionLogic - Happy flow', () => {
+    spies.getRandomId.mockReturnValueOnce('123');
+    spies.isStepFunction.mockReturnValueOnce(true);
+    spies.addStepFunctionEvent.mockImplementationOnce(() => {});
```
doriaviram: Try to don't mock your functions

src/tracer.js
```diff
@@ -131,12 +136,34 @@ const performPromisifyType = (err, data, type, callback) => {
   }
 };
 
+export const performStepFunctionLogic = handlerReturnValue => {
+  return (
+    safeExecute(() => {
+      if (!isStepFunction()) {
+        return handlerReturnValue;
+      }
```
doriaviram: If it's not step function don't call this method, it's safer and cleaner then a function that does nothing in some cases

src/tracer.js
```diff
@@ -131,12 +136,34 @@ const performPromisifyType = (err, data, type, callback) => {
...
+      const messageId = getRandomId();
+
+      addStepFunctionEvent(messageId);
+
+      const { data } = handlerReturnValue;
```
doriaviram: how we sure it have data ?

saartochner: this is the format that `promisifyUserHandler` returns 🤔

src/tracer.test.js
saartochner: The other option is to mock the internals of this function (because it otherwise just doesn't work).
I really believe that it is better to mock the upper level "API" instead of inner implementation...

src/utils.js
saartochner: Let's talk about it, call me :-)

src/tracer.js
saartochner: I'll destruct it in the beginning, and it will be clearer :)

PR #96, opened 2020-03-26
RD-2352 Timeout timer & send http request spans

doriaviram:


src/utils.js
```diff
@@ -126,6 +126,12 @@ export const getAWSEnvironment = () => {
 
 export const isAwsEnvironment = () => !!process.env['LAMBDA_RUNTIME_DIR'];
 
+export const isTimeoutTimerEnabled = () =>
+  !!(
+    process.env['LUMIGO_TIMEOUT_TIMER_ENABLED'] &&
```
nirLumigo: IMO it will be nice to extract `'LUMIGO_TIMEOUT_TIMER_ENABLED'` to a const because it is used in both the `get` and `set` functions.

src/hooks/http.test.js
```diff
@@ -493,6 +502,157 @@ describe('http hook', () => {
...
+
+  test('httpRequestWrapper - added span before request finish', () => {
+    utils.setTimeoutTimerEnabled();
+    TracerGlobals.setHandlerInputs({
+      event: {},
+      context: {
+        awsRequestId: HttpSpanBuilder.DEFAULT_PARENT_ID,
+        invokedFunctionArn: HttpSpanBuilder.DEFAULT_ARN,
+        functionVersion: HttpSpanBuilder.DEFAULT_VERSION,
+      },
```
nirLumigo: Maybe extract this to a variable or a function, it is used in more than one test.

src/hooks/http.test.js
```diff
@@ -493,6 +502,157 @@ describe('http hook', () => {
...
+    });
+    const cleanRequestData = HttpSpanBuilder.DEFAULT_REQUEST_DATA;
+    let a = {};
+    let requestData = { ...cleanRequestData };
```
nirLumigo: Why do we need `requestData` not just use `cleanRequestData`? The object should be circular so `cleanRequestData` might even be the correct choice.

src/hooks/http.test.js
```diff
@@ -493,6 +502,157 @@ describe('http hook', () => {
...
+        request: requestData,
+        response: responseData,
+      })
+      .withRequestTimesFromSpan(spans[0])
+      .build();
+
+    expect(spans).toEqual([expectedSpan]);
+  });
+
+  test('httpRequestWrapper - circular object', () => {
```
nirLumigo: Add 2 words explaining what should happen with a circular object.

src/hooks/http.test.js
```diff
@@ -545,26 +705,13 @@ describe('http hook', () => {
...
-      type: 'http',
-      info: {
-        resourceName: 'StepFunction',
-        httpInfo: { host: 'StepFunction' },
-        messageId: '123',
-        additional: 'bla',
-      },
-      started: 123,
-    });
+    const spans = SpansContainer.getSpans();
```
nirLumigo: Much better :)

src/spans/awsSpan.test.js
```diff
@@ -622,7 +619,78 @@ describe('awsSpan', () => {
...
+          Root: '1-5cdcf03a-64a1b06067c2100c52e51ef4',
+          Sampled: '0',
+          transactionId: '64a1b06067c2100c52e51ef4',
+        },
+        tracer: {
+          name: '@lumigo/tracerMock',
+          version: '1.2.3',
+        },
+      },
+      memoryAllocated: '1024',
```
nirLumigo: Many of these keys are coming from `SpanBuilder`. Do you think they should be part of the test?

src/spans/awsSpan.js
```diff
@@ -206,38 +205,49 @@ export const getHttpInfo = (requestData, responseData) => {
...
 
 export const getHttpSpanTimings = (requestData, responseData) => {
   const { sendTime: started } = requestData;
-  const { receivedTime: ended } = responseData;
+  const { receivedTime: ended } = responseData || {};
   return { started, ended };
 };
 
-export const getHttpSpan = (requestData, responseData) => {
+export const getHttpSpanId = (randomRequestId, awsRequestId = null) => {
```
nirLumigo: IMO the name is a bit strange, when calling `getHttpSpanId` I wouldn't expect to pass the `id` to the function.

testUtils/httpsMocker.js
```diff
@@ -13,11 +14,21 @@ export const HttpsScenarioBuilder = (() => {
...
   };
 
-  const getNextResponse = () => {
+  const getNextData = () => {
     const response = nextResponse.pop();
     return response ? response : defaultResponse;
   };
 
+  const isRequestShouldFinish = () => {
+    if (isNextRequestShouldFinish) return true;
```
nirLumigo: Shouldn't the value of `isNextRequestShouldFinish` be updated?

package.json
```diff
@@ -48,7 +48,9 @@
     "semantic-release": "^15.13.19",
     "shimmer": "^1.2.1",
     "webpack": "^4.35.0",
-    "webpack-cli": "^3.3.4"
+    "webpack-cli": "^3.3.4",
+    "@babel/plugin-proposal-class-properties": "^7.8.3",
```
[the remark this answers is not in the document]
doriaviram: Its to use classes on our tests, not it's not added to the dist, (also its under devDeps)

src/hooks/http.js
```diff
@@ -219,25 +232,34 @@ export const httpRequestWrapper = originalRequestFn =>
         const traceId = getPatchedTraceId(awsXAmznTraceId);
         options.headers['X-Amzn-Trace-Id'] = traceId;
       }
-
       const requestData = parseHttpRequestOptions(options, url);
+      const requestRandomId = getRandomId();
+
+      if (isTimeoutTimerEnabled()) {
```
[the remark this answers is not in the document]
doriaviram: We want that this feature will only for specific customers, this feature relevant mostly for timeout cases (to see clearly which HTTP Call cause the timeout, in other cases we send all span anyway)

src/hooks/http.js
```diff
@@ -255,7 +277,7 @@ export const httpGetWrapper = httpModule => (/* originalGetFn */) =>
   };
 
 export const addStepFunctionEvent = messageId => {
-  const httpSpan = getHttpSpan({}, {});
+  const httpSpan = getHttpSpan(messageId, {});
```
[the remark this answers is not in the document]
doriaviram: You object with that string ? not sure I understand, why you think its better?

src/utils.js
doriaviram: 👍 
also did there some nice DRY Change take a look

src/hooks/http.test.js
doriaviram: Yes you right, added a builder for that. tell me what you think

src/hooks/http.test.js
doriaviram: After the test, I want to assert the span to be same as the clean data (i want to test that my code remove the circles

src/hooks/http.test.js
doriaviram: We are getting the headers as JSON from our customers, they can do whatever they want and all standards HTTP client support this shit, so we need also to not cursing on stack overflow (This test added with blood ;) )

src/hooks/http.test.js
doriaviram: 👑

src/spans/awsSpan.test.js
doriaviram: I agree in general, this test specify checks the build of the spans, I don't want to hide bug under our builder

src/spans/awsSpan.js
doriaviram: this function just decides to take awsRequestId instead to do it inline, how you suggest calling it?

testUtils/httpsMocker.js
doriaviram: No, the default state of isNextRequestShouldFinish is `true` so we don't want to change it.
if its false we get this back to a `true`

src/hooks/http.test.js
doriaviram: Fixed

src/hooks/http.test.js
doriaviram: Added explain

PR #98, opened 2020-04-01
RD-2352 Timeout Timer

doriaviram:


src/globals.js
```diff
@@ -13,6 +13,22 @@ export const SpansContainer = (() => {
   return { addSpan, getSpans, clearSpans };
 })();
 
+export const GlobalTimer = (() => {
+  let currentTimer = undefined;
+
+  const setGlobalTimeout = (func, duration) => {
+    clearTimer();
+    currentTimer = setTimeout(func, duration);
+    currentTimer.unref();
```
GuyMoses, line 22: so it might never be called right? if the lambda exists somehow before the timeout

src/globals.js
```diff
@@ -86,4 +102,5 @@ export const clearGlobals = () => {
   SpansContainer.clearSpans();
   TracerGlobals.clearTracerInputs();
   TracerGlobals.clearHandlerInputs();
+  GlobalTimer.clearTimer();
```
GuyMoses, line 105: why export is needed here?

src/globals.test.js
```diff
@@ -19,6 +19,68 @@ describe('globals', () => {
...
+        }, ms)
+      );
+
+    globals.GlobalTimer.setGlobalTimeout(async () => {
+      await addToArrayAsync(1);
+      expect(array).toEqual([1]);
+      done();
+    }, 1);
+  });
+
```
GuyMoses, line 70: maybe worth adding a test for async clear?

src/tracer.js
```diff
@@ -43,21 +64,15 @@ export const startTrace = async () => {
...
-        const { rtt } = await sendSingleSpan(functionSpan);
-        return addRttToFunctionSpan(functionSpan, rtt);
-      } else {
-        logger.debug(
-          "Skip sending start span because tracer in 'send only if error' mode ."
-        );
-        return null;
-      }
-    } else {
-      return null;
```
GuyMoses, line 59: so no sendOnlyIfErrors?

src/tracer.test.js
```diff
@@ -52,53 +56,114 @@ describe('tracer', () => {
...
+    const requests = HttpsRequestsForTesting.getRequests();
+    expect(requests.length).toEqual(0);
+  });
 
-    spies.logWarn.mockImplementationOnce(() => {});
-    const err1 = new Error('stam1');
-    spies.isSwitchedOff.mockImplementationOnce(() => {
-      throw err1;
-    });
+  test('startTrace - not sending start-span on SEND_ONLY_ON_ERROR', async () => {
```
GuyMoses, line 91: test also send_only_on_error within timeout case

src/globals.js
doriaviram: Yes, to protect us from the `callbackWaitsForEmptyEventLoop`

src/globals.js
doriaviram: Which export? for clearGlobals?
most of the exports are for testing

src/globals.test.js
doriaviram: 💪 💪 Adding

src/tracer.js
doriaviram: We have protected in `sendSpans` I added also a test to `startTrace` with this case

src/tracer.test.js
doriaviram: 💪 💪  Adding

src/globals.js
nirLumigo: Yes: "If there is a Timeout object scheduled using a set function, unref() can be called on that object. This will change the behavior slightly, and not call the Timeout object if it is the last code to execute. The Timeout object will not keep the process alive, waiting to execute."
https://nodejs.org/uk/docs/guides/timers-in-node/

doriaviram: Yes, Nir you right
This is the exact behavior we want

src/tracer.test.js
```diff
@@ -52,53 +56,137 @@ describe('tracer', () => {
...
+
+    await tracer.startTrace();
+    await tracer.startTrace();
+    SpansContainer.addSpan({ id: 'SomeRandomHttpSpan' });
+
+    setTimeout(() => {
+      const requests = HttpsRequestsForTesting.getRequests();
+      expect(requests.length).toEqual(0);
+      done();
+    }, timeout + testBuffer);
```
nirLumigo, line 189: Same as the test above.

testUtils/environmentBuilder.js
```diff
@@ -0,0 +1,21 @@
...
+  notAwsEnvironment = () => {
+    this._envToDelete.push('LAMBDA_RUNTIME_DIR');
+    return this;
+  };
+
+  applyEnv = () => {
+    process.env = { ...process.env, ...this._envs };
+    this._envToDelete.forEach(env => delete process.env[env]);
+  };
+}
```
nirLumigo, line 21: I think we also need to clean the env after each test.

testUtils/handlerInputesBuilder.js
```diff
@@ -4,13 +4,15 @@ export class HandlerInputesBuilder {
   static DEFAULT_AWS_REQUEST_ID = HttpSpanBuilder.DEFAULT_PARENT_ID;
   static DEFAULT_INVOKED_FUNCTION_ARN = HttpSpanBuilder.DEFAULT_ARN;
   static DEFAULT_FUNCTION_VERSION = HttpSpanBuilder.DEFAULT_VERSION;
+  static DEFAULT_TIMEOUT = 10 * 1000;
```
nirLumigo, line 7: This is only for tests. Right?
If we can't find the timeout for the client we're not putting 10 seconds timer?

src/globals.test.js
```diff
@@ -19,6 +19,81 @@ describe('globals', () => {
...
+  test('GlobalTimer - clears with clearGlobals', done => {
+    const arr = [];
+    globals.GlobalTimer.setGlobalTimeout(() => {
+      arr.push(1);
+    }, 1);
```
nirLumigo, line 75: Maybe use a higher number than 1, so it won't be flaky.

src/globals.test.js
```diff
@@ -19,6 +19,81 @@ describe('globals', () => {
     expect(globals.SpansContainer.getSpans()).toEqual([span2]);
   });
 
+  test('GlobalTimer - simple flow', done => {
+    globals.GlobalTimer.setGlobalTimeout(() => {
```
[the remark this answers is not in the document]
doriaviram: The test not ending until we calling the `done` function, so we test the timer call his function

src/tracer.js
```diff
@@ -29,6 +35,21 @@ export const ASYNC_HANDLER_REJECTED = 'async_handler_rejected';
...
+const setupTimeoutTimer = () => {
+  logger.debug('Timeout timer set-up started');
+  const { context } = TracerGlobals.getHandlerInputs();
+  const { remainingTimeInMillis } = getContextInfo(context);
+  if (TIMEOUT_BUFFER_MS < remainingTimeInMillis) {
+    GlobalTimer.setGlobalTimeout(async () => {
+      logger.debug('The tracer reached the end of the timeout timer');
+      const spans = SpansContainer.getSpans();
+      SpansContainer.clearSpans();
+      await sendSpans(spans);
```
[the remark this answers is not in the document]
doriaviram: In this way, I created "lock", on `await` he releases the event loop and run the next function (that can be `endTrace` that also send spans if we will clear after sending its can cause a race condition.
In this way, we create "atomic" `sendSpans`

src/tracer.test.js
```diff
@@ -52,53 +56,137 @@ describe('tracer', () => {
...
+      .withTimeout(timeout)
+      .build();
+    TracerGlobals.setHandlerInputs(handlerInputs);
+
+    await tracer.startTrace();
+    SpansContainer.addSpan({ id: 'SomeRandomHttpSpan' });
+
+    setTimeout(() => {
+      const requests = HttpsRequestsForTesting.getRequests();
+      expect(requests.length).toEqual(2);
```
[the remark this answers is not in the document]
doriaviram: You right, its mistake, removeing

src/tracer.test.js
```diff
@@ -52,53 +56,137 @@ describe('tracer', () => {
...
+    TracerGlobals.setHandlerInputs(handlerInputs);
+
+    await tracer.startTrace();
+    await tracer.startTrace();
+    SpansContainer.addSpan({ id: 'SomeRandomHttpSpan' });
+
+    setTimeout(() => {
+      const requests = HttpsRequestsForTesting.getRequests();
+      //Expect 3 - 2 start spans and 1 from the timer
+      expect(requests.length).toEqual(3);
```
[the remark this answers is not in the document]
doriaviram: The startTrace any way send start span (its okay in our system, we know to handle this dups), the test here checks that in any case we only open 1 timer

src/globals.js
```diff
@@ -13,6 +13,22 @@ export const SpansContainer = (() => {
...
+  let currentTimer = undefined;
+
+  const setGlobalTimeout = (func, duration) => {
+    clearTimer();
+    currentTimer = setTimeout(func, duration);
+    currentTimer.unref();
+  };
+
+  const clearTimer = () => {
+    clearTimeout(currentTimer);
```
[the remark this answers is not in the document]
doriaviram: No, its handle this well, I will add a test for this case anyway, good point :)

src/tracer.test.js
```diff
@@ -52,53 +56,137 @@ describe('tracer', () => {
...
+    utils.setTimeoutTimerEnabled();
+    utils.setSendOnlyIfErrors();
+    new EnvironmentBuilder().awsEnvironment().applyEnv();
+    const handlerInputs = new HandlerInputesBuilder()
+      .withTimeout(timeout)
+      .build();
+    TracerGlobals.setHandlerInputs(handlerInputs);
+
+    await tracer.startTrace();
+    await tracer.startTrace();
```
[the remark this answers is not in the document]
doriaviram: You right, fixed

src/tracer.test.js
```diff
@@ -52,53 +56,137 @@ describe('tracer', () => {
...
+
+    setTimeout(() => {
+      const requests = HttpsRequestsForTesting.getRequests();
+      //Expect 3 - 2 start spans and 1 from the timer
+      expect(requests.length).toEqual(3);
+      done();
+    }, timeout + testBuffer);
+  });
+
+  test('startTrace - timeout timer - to short timeout (timer not effects)', async done => {
```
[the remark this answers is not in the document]
doriaviram: No, it's why I'm checking that is only 1 timer running at the time

src/tracer.test.js
```diff
@@ -52,53 +56,137 @@ describe('tracer', () => {
...
+
+    await tracer.startTrace();
+    SpansContainer.addSpan({ id: 'SomeRandomHttpSpan' });
+
+    setTimeout(() => {
+      const requests = HttpsRequestsForTesting.getRequests();
+      //Expect 1 for start span
+      expect(requests.length).toEqual(1);
+      done();
+    }, timeout + testBuffer);
```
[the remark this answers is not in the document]
doriaviram: its 10ms, I don't want the test to be flaky

src/tracer.test.js
doriaviram: its 10ms, I don't want the test to be flaky

testUtils/environmentBuilder.js
doriaviram: We make this already in `jest.setup.js`

testUtils/handlerInputesBuilder.js
doriaviram: Yes

src/globals.test.js
doriaviram: This is the push, not the test
sooner is better :)

src/tracer.test.js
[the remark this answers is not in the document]
doriaviram: 1 for start span, 1 for the span I added (added comment also)

src/globals.test.js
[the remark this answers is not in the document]
nirLumigo: Maybe add a line explaining that

PR #100, opened 2020-04-05
RD-2352

doriaviram:


src/globals.js
```diff
@@ -1,18 +1,34 @@
...
   return { addSpan, getSpans, clearSpans };
 })();
 
+export const GlobalTimer = (() => {
+  let currentTimer = undefined;
+
+  const setGlobalTimeout = (func, duration) => {
+    clearTimer();
+    currentTimer = setTimeout(func, duration);
+    currentTimer.unref();
```
theburningmonk, line 22: why is this needed? isn't `clearTimer` going to take care of it? is this here "just in case"?

src/hooks/http.js
```diff
@@ -219,25 +232,34 @@ export const httpRequestWrapper = originalRequestFn =>
         const traceId = getPatchedTraceId(awsXAmznTraceId);
         options.headers['X-Amzn-Trace-Id'] = traceId;
       }
-
       const requestData = parseHttpRequestOptions(options, url);
+      const requestRandomId = getRandomId();
```
theburningmonk, line 236: I don't understand the significance of `requestRandomId` in relation to the timer

src/tracer.js
```diff
@@ -29,6 +35,21 @@ export const ASYNC_HANDLER_REJECTED = 'async_handler_rejected';
 export const NON_ASYNC_HANDLER_ERRORED = 'non_async_errored';
 export const LEAK_MESSAGE =
   'Execution leak detected. More information is available in: https://docs.lumigo.io/docs/execution-leak-detected';
+const TIMEOUT_BUFFER_MS = 500;
```
theburningmonk, line 38: half a second is quite a long time, why pick such a high number?

src/tracer.js
```diff
@@ -29,6 +35,21 @@ export const ASYNC_HANDLER_REJECTED = 'async_handler_rejected';
...
+
+const setupTimeoutTimer = () => {
+  logger.debug('Timeout timer set-up started');
+  const { context } = TracerGlobals.getHandlerInputs();
+  const { remainingTimeInMillis } = getContextInfo(context);
+  if (TIMEOUT_BUFFER_MS < remainingTimeInMillis) {
+    GlobalTimer.setGlobalTimeout(async () => {
+      logger.debug('The tracer reached the end of the timeout timer');
```
theburningmonk, line 46: this message is confusing, how about saying something that tells the user what's actually happening? e.g. `Invocation is about to timeout, sending trace data back to Lumigo`

doriaviram: 👌

src/tracer.js
doriaviram: We going with `250ms`, maybe we will reduce this later

PR #101, opened 2020-04-12
RD-2540 print recursive loop

doriaviram:


src/logger.js
```diff
@@ -1,29 +1,68 @@
...
+export const LogStore = (() => {
+  let logSet = new Set([]);
+  let duplicateLogsCount = 0;
+
+  const addLog = (type, message, object) => {
+    const logObj = JSON.stringify({ type, message, object });
+    if (!logSet.has(logObj)) {
+      logSet.add(logObj);
+    } else {
+      duplicateLogsCount++;
```
nirLumigo, line 17: This means that we sum all duplicates instead of counting grouped by duplicate-type.
For example: `"a", "a", "b", "b"`  will be `4` instead of `{a: 2, b:2}`.
Is this the behavior we want?

src/logger.js
```diff
@@ -1,29 +1,68 @@
...
+    }
+    isEmergencyMode() && printLogs();
```
nirLumigo, line 19: I think that our convention is using `if isEmergencyMode()`. Isn't it?

src/logger.js
```diff
@@ -1,29 +1,68 @@
...
+
+  const printLogs = () => {
+    logSet.forEach(logObj => {
+      const { message, obj } = JSON.parse(logObj);
+      forceLog('FATAL', message, obj);
+    });
+    logSet.clear();
+  };
+
+  const isEmergencyMode = () => duplicateLogsCount >= MAX_DUPLICATE_LOGS;
```
nirLumigo, line 30: Do we identify an infinite loop by checking if there are duplicates of logs?
There could be an infinite loop without logs and there could be logs without an infinite loop.
Are we ok with this?

src/logger.js
```diff
@@ -1,29 +1,68 @@
...
+export const fatal = invokeLog('FATAL');
 
-export const fatal = exports.invokeLog('FATAL');
+export const debug = invokeLog('DEBUG');
 
-export const debug = exports.invokeLog('DEBUG');
+const log = (levelname, message, obj) => {
+  const storeLogsIsOn = isStoreLogs();
+  storeLogsIsOn && LogStore.addLog(levelname, message, obj);
+  if (exports.isDebug() && !storeLogsIsOn) {
```
nirLumigo, line 56: This means we will look only at our logs and not the user's logs. Is this ok?

src/logger.test.js
```diff
@@ -1,101 +1,265 @@
...
+    utils.setStoreLogsOn();
+    TracerGlobals.setTracerInputs({});
+
+    logger.info('Test');
+
+    expect(ConsoleWritesForTesting.getLogs()).toEqual([]);
+  });
+
+  test('info -> store logs on -> printing after 50 dups', () => {
+    utils.setDebug();
```
nirLumigo, line 60: Are you sure we need to `setDebug` for this test?
`debug` isn't related to `storeLogs`, isn't it?
Same for other similar tests.

src/utils.js
```diff
@@ -193,6 +196,8 @@ export const setSendOnlyIfErrors = () =>
 export const setPruneTraceOff = () =>
   (process.env[PRUNE_TRACE_OFF_FLAG] = 'TRUE');
 
+export const setStoreLogsOn = () => (process.env[STORE_LOGS_FLAG] = 'TRUE');
```
nirLumigo, line 199: Don't we need a clean function for the tests?

testUtils/consoleMocker.js
```diff
@@ -16,5 +16,9 @@ export const ConsoleWritesForTesting = (() => {
 
 export const ConsoleMocker = (() => {
   const log = ConsoleWritesForTesting.addLog;
-  return { log };
+  const info = ConsoleWritesForTesting.addLog;
+  const debug = ConsoleWritesForTesting.addLog;
+  const warn = ConsoleWritesForTesting.addLog;
+  const error = ConsoleWritesForTesting.addLog;
+  return { log, info, warn, error, debug };
```
nirLumigo, line 23: We have the same set of tests for each log level. It looks like we mock each of the log functions to the same function.
So we actually test the same thing in every test, isn't it?

src/logger.js
doriaviram: Only on frontend maybe, here most of the 1 line ifs it's like that

src/logger.js
doriaviram: Yes, if you have other ideas on how to figure we in an infinite loop you can suggest :)

src/logger.js
doriaviram: Yes.

src/logger.test.js
doriaviram: Yes we can remove this, it doesn't matter

src/utils.js
doriaviram: Not sure I understand, but the test infra here take care of cleaning the env vars each test

testUtils/consoleMocker.js
doriaviram: Ammmm yes, but its black-boxed test
we cant assume they using the same function.

src/logger.js
doriaviram: Yes its good enough for now

PR #107, opened 2020-04-26
fix: add support to apigw v2 [RD-2529]

doriaviram:


src/parsers/aws.js
```diff
@@ -47,6 +47,19 @@ export const snsParser = (requestData, responseData) => {
...
 };
 
+export const apigwParser = (requestData, responseData) => {
+  const baseData = awsParser(requestData, responseData);
+  if (!baseData.awsServiceData) {
+    baseData.awsServiceData = {};
+  }
+
+  const { headers: resHeader } = responseData;
+  if (resHeader['Apigw-Requestid']) {
```
uri-p, line 57: Do you need the if here ?

PR #108, opened 2020-04-26
RD-2640 execution tags

nirLumigo:
Allow the user to add dynamic execution tags.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,8 @@
   "devDependencies": {
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.4",
+    "@babel/plugin-proposal-class-properties": "^7.8.3",
+    "@babel/plugin-proposal-optional-chaining": "^7.9.0",
     "@babel/preset-env": "^7.4.4",
     "@commitlint/cli": "^8.1.0",
     "@commitlint/config-conventional": "^8.0.0",
@@ -38,6 +40,7 @@
     "eslint-config-prettier": "^4.3.0",
     "eslint-plugin-jest": "^22.5.1",
     "eslint-plugin-prettier": "^3.1.0",
+    "fs-extra": "^8.1.0",
     "husky": "^2.4.1",
     "jest": "^24.8.0",
     "jest-junit": "^6.4.0",
@@ -48,9 +51,7 @@
     "semantic-release": "^15.13.19",
     "shimmer": "^1.2.1",
     "webpack": "^4.35.0",
-    "webpack-cli": "^3.3.4",
-    "@babel/plugin-proposal-class-properties": "^7.8.3",
-    "@babel/plugin-proposal-optional-chaining": "^7.9.0"
+    "webpack-cli": "^3.3.4"
   },
   "release": {
     "branch": "master",
--- a/src/globals.test.js
+++ b/src/globals.test.js
@@ -341,7 +341,6 @@ describe('globals', () => {
     });
     globals.ExecutionTags.addTag('throw', 'exception');
     // No exception.
-    let tags = globals.ExecutionTags.getTags();
-    expect(tags).toEqual([]);
+    expect(globals.ExecutionTags.getTags()).toEqual([]);
   });
 });
--- a/src/globals.test.js
+++ b/src/globals.test.js
@@ -284,4 +284,64 @@ describe('globals', () => {
       isStepFunction: false,
     });
   });
+
+  test('ExecutionTags one tag', () => {
+    const value = 'v0';
+    const key = 'k0';
+    globals.ExecutionTags.addTag(key, value);
+    expect(globals.ExecutionTags.getTags()).toEqual([{ key, value }]);
+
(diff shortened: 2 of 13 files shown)


src/index.test.js
```diff
@@ -12,6 +13,43 @@ describe('index', () => {
...
+    const actualTags = sentSpans.filter(
+      span => !span.id.endsWith('_started')
+    )[0][EXECUTION_TAGS_KEY];
+    expect(actualTags).toEqual([
+      { key: 'k0', value: 'v0' },
+      { key: 'k1', value: 'v1' },
+    ]);
+  });
+
+  test('addExecutionTag without tracer not throw exception', async () => {
```
doriaviram, line 47: Test also here the `const lumigo = lumigo_import({ token });` Way

src/utils.js
```diff
@@ -27,6 +27,8 @@ export const LUMIGO_EVENT_KEY = '_lumigo';
 export const STEP_FUNCTION_UID_KEY = 'step_function_uid';
 export const GET_KEY_DEPTH_ENV_KEY = 'LUMIGO_KEY_DEPTH';
 export const DEFAULT_GET_KEY_DEPTH = 3;
+export const EXECUTION_TAGS_KEY = 'lumigo_execution_tags_no_scrub';
```
doriaviram, line 30:
```suggestion
export const EXECUTION_TAGS_KEY = 'lumigoExecutionTags';
```
no_scrub is implementation :)

src/globals.test.js
```diff
@@ -284,4 +284,63 @@ describe('globals', () => {
...
+  test('ExecutionTags.addTag too many tags', () => {
+    for (let i = 0; i < 51; i++) globals.ExecutionTags.addTag(`k${i}`, `v${i}`);
+    expect(globals.ExecutionTags.getTags().length).toEqual(50);
+    expect(
+      globals.ExecutionTags.getTags().filter(tag => tag.key === 'k50')
+    ).toEqual([]);
+  });
+
+  test('ExecutionTags.addTag catch exception', () => {
+    jest.spyOn(global, 'String').mockImplementation(() => {
```
doriaviram, line 339: Try to implement with something like 

```
const obj = {
  get requestContext() {
     throw new Error();
  }
}
```

src/index.js
```diff
@@ -25,5 +26,8 @@ module.exports = function({
       eventFilter,
       stepFunction,
     }),
+    addExecutionTag: ExecutionTags.addTag,
```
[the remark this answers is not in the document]
nirLumigo: In the documentations we use `const lumigo = require('@lumigo/tracer')({ token: 'DEADBEEF' })`.
So it will allow users to do `lumigo.addExecutionTag` without splitting the code above to 2 lines:
```
const lumigoImport = require('@lumigo/tracer')
const lumigo = lumigoImport({ token: 'DEADBEEF' })
lumigoImport.addExecutionTag
```

Efi asked to add this behavior as well.

src/utils.js
nirLumigo: That why I put it in a variable with a normal name :)
There are 2 reasons for this:
1. For the skip-scrubbing mechanism, we want to add only keys that will probably be unique. Also, maybe in the future we'll want to skip scrubbing for all keys that end with "_no_scrub".
2. It's already the key in the python-tracer and in the backend.

src/index.test.js
```diff
@@ -12,6 +13,43 @@ describe('index', () => {
     Object.keys(spies).map(x => spies[x].mockClear());
   });
 
+  test('execution tags', async () => {
```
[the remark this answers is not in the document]
nirLumigo: I'm not 100% sure how it should be implemented in the tests, please take a look at this test.

package.json
```diff
@@ -48,9 +51,7 @@
     "semantic-release": "^15.13.19",
     "shimmer": "^1.2.1",
     "webpack": "^4.35.0",
-    "webpack-cli": "^3.3.4",
-    "@babel/plugin-proposal-class-properties": "^7.8.3",
-    "@babel/plugin-proposal-optional-chaining": "^7.9.0"
+    "webpack-cli": "^3.3.4"
```
doriaviram, line 54: Why we need this ?

src/index.test.js
```diff
@@ -14,16 +15,50 @@ describe('index', () => {
...
+    const originDirPath = __dirname;
+    const dupDirPath = `${originDirPath}Dup'`;
+    const layerPath = `${dupDirPath}/index.js`;
+
+    const context = { getRemainingTimeInMillis: () => 30000 };
+    const callback = jest.fn();
+    const retVal = 'The Tracer Wars';
+
+    try {
+      fsExtra.copySync(originDirPath, dupDirPath);
```
doriaviram, line 28: it's better to use Async and Await

package.json
nirLumigo: It was there before, I just added `fs-extra` to the devDependency

PR #110, opened 2020-04-28
feat: add support for parsing sqs and sns events

uri-p:


src/parsers/eventParser.js
```diff
@@ -84,14 +113,51 @@ export const parseApiGwEvent = event => {
   return parsed_event;
 };
 
+export const parseSnsEvent = event => {
+  const new_sns_event = {};
+  new_sns_event['Records'] = [];
+  // Add order keys
+  for (const rec of event['Records']) {
```
doriaviram, line 120: Create some util func to this code duplicate

doriaviram: Ordering keys by list

uri-p: It has some specific logic in each one of the code, making a generic function will make to code unclear

PR #112, opened 2020-05-05
fix: apigw v2 + better error handling rd-2529

doriaviram:


src/parsers/aws.js
```diff
@@ -47,6 +49,25 @@ export const snsParser = (requestData, responseData) => {
...
+  logger.info('APIGW Parser req', requestData);
+  logger.info('APIGW Parser res', responseData);
+  if (!responseData) return {};
+  const baseData = awsParser(requestData, responseData);
+  if (!baseData.awsServiceData) {
+    baseData.awsServiceData = {};
+  }
+
+  if (!baseData.awsServiceData.messageId) {
+    const { headers: resHeader } = responseData;
```
uri-p, line 62: resHeader might also be none no?

doriaviram: 👍

PR #114, opened 2020-05-24
RD-2901 - override step function key if already exists in return value

saartochner:


src/tracer.test.js
```diff
@@ -529,7 +528,7 @@ describe('tracer', () => {
 
   test('No exception at initialization', async done => {
     const mockedTracerGlobals = jest.spyOn(TracerGlobals, 'setHandlerInputs');
-    mockedTracerGlobals.mockImplementation(() => {
+    mockedTracerGlobals.mockImplementationOnce(() => {
```
saartochner, line 531: changed because it affected the other tests!
Couldn't find way to write this without mocks :(

PR #120, opened 2020-06-02
fix: add node-timer

doriaviram:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lumigo/tracer",
-  "version": "1.35.3",
+  "version": "2.0.28",
```
efimk-lu, line 3: Why?

src/reporter.js
```diff
@@ -82,18 +95,26 @@ export const forgeRequestBody = (spans, maxSendBytes = MAX_SENT_BYTES) => {
   );
 
   const orderedSpans = [...errorSpans, ...normalSpans];
+  TracerTimer.endJob('spanHasErrors');
 
+  TracerTimer.startJob('getJSONBase64Size');
```
efimk-lu, line 100: No need to keep these lines, right

PR #122, opened 2020-06-09
RD-3002 - add dynamodb-stream support

saartochner:


src/events.js
```diff
@@ -73,11 +74,42 @@ export const getKinesisData = event => {
...
+        record.eventName === 'MODIFY' &&
+        record.dynamodb &&
+        record.dynamodb.Keys
+      ) {
+        return md5Hash(record.dynamodb.Keys);
+      } else if (
+        record.eventName === 'REMOVE' &&
+        record.dynamodb &&
+        record.dynamodb.Keys
+      ) {
```
doriaviram, line 93: DRY

src/events.js
```diff
@@ -73,11 +74,42 @@ export const getKinesisData = event => {
...
+        return md5Hash(record.dynamodb.Keys);
+      } else if (
+        record.eventName === 'INSERT' &&
+        record.dynamodb &&
+        record.dynamodb.NewImage
+      ) {
+        return md5Hash(record.dynamodb.NewImage);
+      }
+    })
+    .filter(x => !!x);
```
doriaviram, line 103: Why we need this ?

saartochner: if one of the records is not INSERT, MODIFY, REMOVE then it will be undefined - so we want to remove it from the list

PR #127, opened 2020-06-14
feat: add support to got http lib [RD-3047]

doriaviram:


.circleci/config.yml
```diff
@@ -233,8 +233,6 @@ workflows:
         filters:
           branches:
             ignore: master
-        requires:
-        - test
```
doriaviram, line 237: Ignore this part, not going to push this.
It's just, for now, I want the integration will run. before implementing all UT

src/hooks/http.js
```diff
@@ -68,14 +68,35 @@ export const parseHttpRequestOptions = (options = {}, url) => {
...
+    body: '', // XXX Filled by the httpRequestEndWrapper or httpRequestEmitWrapper ( / Write)
     method,
     headers: lowerCaseObjectKeys(headers),
     protocol,
     sendTime,
   };
 };
 
+export const httpRequestEmitWrapper = requestData => emitFunc =>
+  function(eventName, arg) {
```
nirLumigo, line 80: Why isn't it an arrow function like all the others?

src/hooks/http.js
```diff
@@ -68,14 +68,35 @@ export const parseHttpRequestOptions = (options = {}, url) => {
...
+      () => {
+        if (arg && arg._httpMessage && arg._httpMessage._hasBody) {
+          const httpMessage = arg._httpMessage;
+          let lines = [];
+          if (httpMessage.hasOwnProperty('outputData')) {
+            lines = httpMessage.outputData[0].data.split('\n');
+          } else if (httpMessage.hasOwnProperty('output')) {
+            lines = httpMessage.output[0].split('\n');
+          }
+          requestData.body += lines[lines.length - 1];
```
nirLumigo, line 91: Are we adding lines to the body of the request?

src/hooks/http.js
```diff
@@ -159,16 +182,16 @@ export const httpRequestArguments = args => {
   let options = undefined;
   let callback = undefined;
 
-  if (typeof args[0] === 'string') {
+  if (typeof args[0] === 'string' || args[0] instanceof URL) {
```
nirLumigo, line 185: Why `instanceof` and not `typeof`?

src/hooks/http.test.js
```diff
@@ -296,7 +389,39 @@ describe('http hook', () => {
...
+    const options = { a: 'b' };
+    const expected1 = {
+      url,
+      options,
+      callback: undefined,
+    };
+    expect(httpHook.httpRequestArguments([url, options])).toEqual(expected1);
+  });
+
+  test('httpRequestArguments -> http(objectUrl, callback)', () => {
```
nirLumigo, line 413: Shouldn't there be also a test with `http(callback)`?

src/hooks/http.js
```diff
@@ -68,14 +68,35 @@ export const parseHttpRequestOptions = (options = {}, url) => {
...
 };
 
+export const httpRequestEmitWrapper = requestData => emitFunc =>
+  function(eventName, arg) {
+    safeExecute(
+      () => {
+        if (arg && arg._httpMessage && arg._httpMessage._hasBody) {
+          const httpMessage = arg._httpMessage;
+          let lines = [];
+          if (httpMessage.hasOwnProperty('outputData')) {
```
saartochner, line 86: validate that you can run `httpMessage.outputData[0].data.split`, i.e.  that data is string, etc.

src/hooks/http.js
saartochner, line 91: why do you take only the last line?
Don't you want to take all the lines to the body?

src/hooks/http.js
```diff
@@ -267,12 +292,20 @@ export const httpRequestWrapper = originalRequestFn =>
         logger.warn('end wrap error', e.message);
       }
 
+      if (!isRequestToAwsService) {
```
saartochner, line 295: huge smell-
Can you share what happened when you did it to aws requests?

saartochner: If it's destructive, then we might destruct customer's requests...

src/hooks/http.js
```diff
@@ -267,12 +292,20 @@ export const httpRequestWrapper = originalRequestFn =>
...
+        try {
+          const emitWrapper = httpRequestEmitWrapper(requestData);
+          shimmer.wrap(clientRequest, 'emit', emitWrapper);
```
saartochner, line 298: you're adding another shimmering - can you stress test that?

src/hooks/http.js
saartochner: Moreover, if lines is empty, then you mess `requestData.body` with `+=undefined` which cast it to string 😂

saartochner:
I'm missing an "e2e" test that uses the `emit` functionality and checks that we collected the right data

src/hooks/http.js
doriaviram: It's wrapped in `safeExecute`

src/hooks/http.js
doriaviram: Yes, I'm adding in integration-test

src/hooks/http.js
doriaviram: Fixed :)

src/hooks/http.test.js
doriaviram: Didn't find this case in the docs

src/index.test.js
```diff
@@ -15,6 +15,7 @@ describe('index', () => {
     Object.keys(spies).map(x => spies[x].mockClear());
   });
 
+  jest.setTimeout(15000);
```
saartochner, line 18: put it inside the relevant test (and not on the class level)

src/utils.js
```diff
@@ -497,3 +497,17 @@ export const md5Hash = item => {
   md5sum.update(JSON.stringify(item));
   return md5sum.digest('hex');
 };
+
+export const isEncodingType = encodingType =>
+  !!(
+    encodingType &&
+    typeof encodingType === 'string' &&
+    ['ascii', 'utf8', 'utf16le', 'ucs2', 'base64', 'binary', 'hex'].includes(
```
saartochner, line 505: there's also `utf-8`, `utf-16-le` etc.:
https://stackoverflow.com/questions/14551608/list-of-encodings-that-node-js-supports
maybe you can import this list from somewhere else?

src/utils.js
```diff
@@ -497,3 +497,17 @@ export const md5Hash = item => {
...
+      encodingType
+    )
+  );
+
+export const isEmptyString = str => !!(!str || str.length === 0);
```
saartochner, line 510: add `typeof` to avoid `isEmptyString([])`

src/utils.js
```diff
@@ -497,3 +497,17 @@ export const md5Hash = item => {
...
+
+export const isValidHttpRequestBody = reqBody =>
+  !!(reqBody && (typeof reqBody === 'string' || reqBody instanceof Buffer));
```
saartochner, line 513: empty string is not valid?

src/hooks/http.test.js
```diff
@@ -130,6 +130,172 @@ describe('http hook', () => {
     expect(requestData).toEqual({ body: '' });
   });
 
+  test('httpRequestWriteWrapper -> simple flow -> write(str)', dontExitWithoutDone => {
+    const requestData = {
+      body: '',
+    };
+    const wrapper = httpHook.httpRequestWriteWrapper(requestData);
```
saartochner, line 137: can you lower this line to where you use it?
It'll help the reader...

src/index.test.js
doriaviram: I will move this, but Its doesn't matter in fact

src/hooks/http.js
```diff
@@ -146,7 +169,7 @@ export const wrappedHttpResponseCallback = (
 
 export const httpRequestEndWrapper = requestData => originalEndFn =>
   function(...args) {
-    if (args[0] && (typeof args[0] === 'string' || args[0] instanceof Buffer)) {
```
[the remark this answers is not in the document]
doriaviram: 👍

src/utils.js
doriaviram: Its buffer encodings found this list here
https://stackoverflow.com/questions/34476754/what-encodings-does-buffer-tostring-support

src/utils.js
doriaviram: 👍

src/utils.js
doriaviram: No, it's mean its unfilled

src/hooks/http.test.js
doriaviram: 👍

PR #128, opened 2020-06-16
feat: take the approximate creation time of dynamodb triggers in ms

saartochner:


src/utils.js
```diff
@@ -492,7 +493,12 @@ const recursiveGetKeyByDepth = (event, keyToSearch, maxDepth) => {
 };
 
 export const md5Hash = item => {
-  const md5sum = crypto.createHash('md5');
-  md5sum.update(JSON.stringify(item));
-  return md5sum.digest('hex');
+  try {
```
doriaviram, line 496: You can use `safeExecute`

saartochner: That's not about the "safeness" but about don't care if we failed 🤷

PR #133, opened 2020-06-29
RD-3124 - fix: omit keys before prune strings

saartochner:
The problem: we don't omit keys of pruned data because it's not a legal json
Solution: omit keys before prune


saartochner:
Not clear that this is the right solution,
checkout https://lumigo.atlassian.net/browse/RD-3124

PR #135, opened 2020-07-05
RD-3162 - follow AWS's return logic

saartochner:


src/spans/awsSpan.js
```diff
@@ -134,9 +134,21 @@ export const removeStartedFromId = id => id.split('_')[0];
...
+  } catch (e) {
+    return_value = data.toString();
+    error = {
+      type: 'ReturnValueError',
+      message: `Could not JSON.stringify the return value. This will probably fail the lambda. Original error: ${e &&
+        e.message}`,
+      stacktrace: {},
+    };
+  }
+
```
doriaviram: Extract this logic to a function

PR #138, opened 2020-07-14
fix: allow to config timeout timer buffer

doriaviram:


src/utils.test.js
```diff
@@ -323,6 +323,46 @@ describe('utils', () => {
     process.env['LUMIGO_VALID_ALIASES'] = undefined;
   });
 
+  test('getTimeoutTimerBuffer -> ENV_VAR', () => {
+    process.env['LUMIGO_TIMEOUT_BUFFER'] = '0.35';
+    expect(utils.getTimeoutTimerBuffer()).toEqual(350);
```
saartochner, line 328: how is that possible that it lower than 500? 😂

src/utils.test.js
```diff
@@ -323,6 +323,46 @@ describe('utils', () => {
...
+  test('getTimeoutTimerBuffer -> ENV_VAR ms', () => {
+    process.env['LUMIGO_TIMEOUT_BUFFER_MS'] = '350';
+    expect(utils.getTimeoutTimerBuffer()).toEqual(350);
+  });
+
+  test('getTimeoutTimerBuffer -> Min value', () => {
+    expect(utils.isValidAlias()).toEqual(true);
+    TracerGlobals.setHandlerInputs({
+      context: {
+        getRemainingTimeInMillis: () => 1000,
```
saartochner, line 340: Just to make sure: you have something that limit the TO mechanism to not work on these short lambdas, right?

src/utils.test.js
doriaviram: It's from the config, its override our mechanism

src/utils.test.js
doriaviram: Only if the getRemainingTimeInMillis < Timeout buffer. 
What the minimum should be ?

saartochner: That's a problem because if the lambda run 600ms, then the TO will jump 100ms after it started... :(
In the python tracer we added a minimum of 2s to the remaining time

doriaviram: Added

PR #142, opened 2020-07-22
fix: keep-alive [RD-3247]

doriaviram:


src/httpSpansAgent.js
```diff
@@ -1,55 +1,66 @@
...
-        baseURL: url,
-        timeout: 250,
-        maxRedirects: 0,
-        HttpSpansAgent: new http.Agent({ keepAlive: true, keepAliveMsecs }),
-        httpsAgent: new https.Agent({ keepAlive: true, keepAliveMsecs }),
-        headers,
-        validateStatus,
+        ...baseConfiguration,
+        httpsAgent: httpAgent,
+        httpAgent: httpAgent,
```
nirLumigo, line 55: They are the same, is this ok?

src/utils.js
```diff
@@ -183,6 +184,8 @@ export const setHttpWrapped = () => (process.env[HTTP_WRAPPED_FLAG] = 'TRUE');
 
 export const isStoreLogs = () => validateEnvVar(STORE_LOGS_FLAG);
 
+export const isReuseHttpConnection = () => validateEnvVar(REUSE_CONNECTION);
```
nirLumigo, line 187: Don't we want the default to be `true`?

src/utils.test.js
```diff
@@ -214,6 +214,12 @@ describe('utils', () => {
...
 
+  test('isReuseHttpConnection', () => {
+    expect(utils.isReuseHttpConnection()).toBe(false);
+    process.env['LUMIGO_REUSE_HTTP_CONNECTION'] = 'TRUE';
+    expect(utils.isReuseHttpConnection()).toBe(true);
+  });
+
   test('setSendOnlyIfErrors', () => {
     expect(utils.isSendOnlyIfErrors()).toBe(false);
     utils.setSendOnlyIfErrors();
```
nirLumigo, line 225: Test idea: wrap a function with the tracer, run it twice, and make sure that only one connection was created.

nirLumigo, line 225: Another issue, if the lambda tries to reuse the connection but the edge closed the connection. What will happen? Will it recreate the connection or the sending of the span will fail?
Unittest: Try to mock closed connection (see how hard it is, it might be too complicated).
Manual test or IT: maybe run a stress test to see that spans are still sent.

src/utils.js
doriaviram: In the feature yes, for now, we want to test this internally first (on backend-monitoring) and then release GA

src/httpSpansAgent.js
doriaviram: Yes :)

src/utils.test.js
doriaviram: Yes, it should re-create it. I'm trying to add a test as u suggested

PR #144, opened 2020-07-22
RD-3178 connection timeout

doriaviram:


src/httpSpansAgent.test.js
```diff
@@ -85,4 +87,20 @@ describe('HttpSpansAgent', () => {
...
+    jest.spyOn(axios, 'create').mockImplementationOnce(() => {
+      return {
+        post: async (url, body, options) => {
+          expect(options.cancelToken).toBeDefined();
+          await sleep(500);
+        },
+      };
+    });
+
+    await HttpSpansAgent.postSpans(reqBody);
```
nirLumigo, line 104: Make sure that `axios. create. post` was called.
You can do it with a variable being updated in the function or with jest.

testUtils/sleep.js
```diff
@@ -0,0 +1,3 @@
+export const sleep = time => {
+  return new Promise(resolve => setTimeout(resolve, time));
+};
```
nirLumigo, line 3: You can add a small test for this.

PR #149, opened 2020-07-28
fix: add better connection timeout support [RD-3178]

doriaviram:


src/httpSpansAgent.js
```diff
@@ -69,18 +72,34 @@ export const HttpSpansAgent = (() => {
...
+          `Edge connection timeout [${CONNECTION_TIMEOUT}ms] from setTimeout (Tracer skipping)`
+        );
+        resolve(1);
+      }, CONNECTION_TIMEOUT);
+    });
+    const requestPromise = session.post(url, requestBody, { headers }).catch(e => {
+      logger.debug('Edge error (Tracer skipping)', e.message);
+    });
+
+    await Promise.race([requestPromise, requestTimeout]).finally(() => {
```
saartochner, line 91: isn't that will leave something in the event loop - which will cause the lambda to not finish? (in the case that the relevant flag is on)

saartochner: why don't we use a package that gives better timeout handling ?

src/httpSpansAgent.test.js
```diff
@@ -84,4 +87,24 @@ describe('HttpSpansAgent', () => {
...
+  test(
+    'postSpans - reject connection timout',
+    async () => {
+      const reqBody = 'abcdefg';
+      globals.TracerGlobals.setTracerInputs({ token: 't_xxx' });
+
+      jest.spyOn(axios, 'create').mockImplementationOnce(() => {
+        return {
+          post: async () => {
+            await sleep(60000);
```
saartochner, line 101: I think that the first line in this function should be `done()` (to validate that it has been called)

doriaviram: We cant use done here, but I'm added that the post really called :)

src/httpSpansAgent.js
doriaviram: Axios timeout handling is not good enough (and is considered as the best), And about the event loop, its a real edge case that our edge will take more than 300ms, it's okay to leave it

saartochner: I disagree here-
if we will do a bad deployment that will crush the edge (or make it to be super slow)- then you may fill the customer's event loop with our events. So if his lambda has many warm requests, then we will kill his production :-(

saartochner: I disagree - we always should consider the case that the edge misbehaves (this is the original cause of this ticket!).
If the lambda will have many warm requests then you'll fill the event loop.. :(

PR #164, opened 2020-09-06
fix: load user modules from layer [RD-3577]

doriaviram:


saartochner (approved):
Isn't that the default in th environment?

PR #170, opened 2020-10-12
feat: double event and envs size limit on error

nirLumigo:


src/spans/awsSpan.js
```diff
@@ -109,8 +121,9 @@ export const getFunctionSpan = () => {
   const started = new Date().getTime();
   const ended = started; // Indicates a StartSpan.
 
-  const event = payloadStringify(parseEvent(lambdaEvent));
-  const envs = payloadStringify(process.env);
+  // We need to keep sending them in the startSpan because we don't always have an endSpan
+  addEventToSpan(basicSpan);
+  addEnvsToSpan(basicSpan);
```
doriaviram, line 126: It's better than its build-in 1 step, we will makespan frozen soon after we will add typescript, so please keep this at the way it was

doriaviram: Just take care of error stuff

src/spans/awsSpan.js
```diff
@@ -156,6 +167,10 @@ export const getEndFunctionSpan = (functionSpan, handlerReturnValue) => {
     return_value,
     [EXECUTION_TAGS_KEY]: ExecutionTags.getTags(),
   });
+  if (error !== undefined) {
+    addEventToSpan(newSpan, true);
+    addEnvsToSpan(newSpan, true);
+  }
```
doriaviram, line 173: Also here, try to build the new span in "one step" and not build and then extend

doriaviram (changes requested):

src/spans/awsSpan.js
```diff
@@ -149,12 +160,16 @@ export const getEndFunctionSpan = (functionSpan, handlerReturnValue) => {
         e.message}`,
     });
   }
+  const event = error === undefined ? functionSpan.event : getEventForSpan(true);
```
doriaviram, line 163:
```suggestion
  const event = error ? getEventForSpan(true) : functionSpan.event;
```

src/spans/awsSpan.js
```diff
@@ -149,12 +160,16 @@ export const getEndFunctionSpan = (functionSpan, handlerReturnValue) => {
...
+  const envs = error === undefined ? functionSpan.envs : getEnvsForSpan(true);
```
doriaviram, line 164:
```suggestion
  const envs = error ? getEnvsForSpan(true) : functionSpan.envs ;
```

doriaviram (approved):

PR #173, opened 2020-10-14
feat: double http payload size on failure

nirLumigo:


src/spans/awsSpan.js
```diff
@@ -270,15 +273,17 @@ export const getHttpSpan = (randomRequestId, requestData, responseData = null) =
   const { awsServiceData, spanId } = serviceData;
 
   const prioritizedSpanId = getHttpSpanId(randomRequestId, spanId);
-  let httpInfo = {
-    host: requestData.host,
-    request: payloadStringify(requestData),
-    response: payloadStringify(responseData),
-  };
```
nirLumigo, line 277: I moved it to be in the `catch` to improve performance (not related to my ticket), not need to to run `payloadStringify` on the http data twice.

doriaviram (approved):

PR #174, opened 2020-10-14
fix: better handling in execution leaks

doriaviram:


src/spans/awsSpan.js
```diff
@@ -56,12 +56,17 @@ export const getCurrentTransactionId = () => {
   return getSpanInfo(lambdaEvent).traceId.transactionId;
 };
 
-export const getBasicSpan = () => {
+export const isSpanIsFromAnotherInvocation = span => {
+  return (
+    !span.id.includes(span.reporterAwsRequestId) && span.parentId !== span.reporterAwsRequestId
```
nirLumigo, line 61: About `span.parentId !== span.reporterAwsRequestId`.
What if there is `_started` in one of the ids? Maybe we should use `includes`?

src/spans/awsSpan.js
```diff
@@ -239,13 +245,13 @@ export const getHttpInfo = (requestData, responseData) => {
   return { host, request, response };
 };
 
-export const getBasicChildSpan = (spanId, spanType) => {
+export const getBasicChildSpan = (transactionId, awsRequestId, spanId, spanType) => {
```
nirLumigo, line 248: Maybe this function should calculate `transactionId` and `awsRequestId` by itself?
It will make the other code simpler and anyway this function access global variables (`TracerGlobals.getHandlerInputs()`).
I wrote a similar comment at the beginning.

src/tracer.js
```diff
@@ -76,22 +77,23 @@ export const startTrace = async () => {
   }
 };
 
+const logLeakedSpans = allSpans => {
+  const warnClientOnce = runOneTimeWrapper(logger.warnClient);
+  allSpans.forEach(span => {
+    if (isSpanIsFromAnotherInvocation(span)) {
+      logger.debug('Leaked span: ', span);
+      warnClientOnce(LEAK_MESSAGE);
```
nirLumigo, line 85: This is in `forEach` but it is wrapped by `runOneTimeWrapper`

src/tracer.test.js
```diff
@@ -525,7 +526,16 @@ describe('tracer', () => {
 
     TracerGlobals.setTracerInputs({ token: '123' });
     spies.SpansContainer.getSpans.mockReturnValueOnce([
-      { transactionId: '123', id: '1' },
+      { transactionId: '123', id: '1', reporterAwsRequestId: '2', parentId: '2' },
+      { transactionId: '456', id: '2' },
+    ]);
+    await tracer.sendEndTraceSpans({ id: '1_started' }, { err: null, data: null });
+    expect(spies.warnClient).not.toHaveBeenCalled();
+    expect(TracerGlobals.getTracerInputs().token).toEqual('');
```
nirLumigo, line 534: What do we test in this `expect`?

nirLumigo (changes requested):
Nice! :)

src/hooks/http.js
```diff
@@ -210,20 +238,31 @@ export const httpBeforeRequestWrapper = (args, extenderContext) => {
...
   if (!isTracedDisabled) {
     const endWrapper = httpRequestEndWrapper(requestData, requestRandomId);
     extender.hook(clientRequest, 'end', { beforeHook: endWrapper });
 
-    const emitWrapper = httpRequestEmitBeforeHookWrapper(requestData, requestRandomId);
+    const emitWrapper = httpRequestEmitBeforeHookWrapper(
+      transactionId,
+      awsRequestId,
+      requestData,
+      requestRandomId
```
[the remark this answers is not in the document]
doriaviram: We need typescript for that, no comments :)

package.json
```diff
@@ -14,7 +14,7 @@
     "build:babel": "babel --ignore '**/*.test.js' --ignore testdata src -d lib --verbose",
     "build:webpack": "webpack --display-modules ./lib -o dist/lumigo.js",
     "build": "npm run build:babel && npm run build:webpack",
-    "jest": "jest",
+    "jest": " jest",
```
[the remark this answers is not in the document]
doriaviram: 👍

src/hooks/mySql.js
```diff
@@ -25,6 +25,8 @@ const createResultHook = currentSpan => args => {
 };
 
 function queryBeforeHook(args, extenderContext) {
+  const awsRequestId = TracerGlobals.getHandlerInputs().context.awsRequestId;
```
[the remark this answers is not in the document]
doriaviram: the `getBasicChildSpan` can be called in the next invocation :)

src/spans/awsSpan.js
doriaviram: The parentId is the request-id, not the span id, so we don't need to include there

src/spans/awsSpan.js
doriaviram: Its need to be calculated when the request is fired, there is a big impact to when the calculation is made, its cant be calculated inside getBasicChildSpan because this function called when the request ends

src/tracer.js
doriaviram: Yes, we want to warn the client just once per invocation, it's instead of the counter-loop that was before

src/tracer.test.js
doriaviram: That the global is cleaned

src/logger.js
```diff
@@ -72,12 +72,15 @@ const forceLog = (levelname, message, obj) => {
   }
 };
 
-export const warnClient = msg => {
+export const warnClient = (msg, obj) => {
```
nirLumigo, line 75:
```suggestion
export const warnClient = (msg, obj = null) => {
```
If someone doesn't want to pass `obj`.

nirLumigo (approved):
Nice!

src/logger.js
doriaviram: Its will be undefined, its ok

PR #183, opened 2020-11-03
feat: adding AppSync triggeredBy

GuyMoses:


src/events.js
```diff
@@ -35,6 +40,17 @@ export const getTriggeredBy = event => {
   return 'invocation';
 };
 
+export const isAppSyncEvent = event => {
+  return (
```
doriaviram, line 44: -1 Will return allways true

doriaviram (approved):

PR #197, opened 2020-12-24
feat(spans): add neo4j support

bboure:
This PR adds support for the [neo4j-driver](https://github.com/neo4j/neo4j-javascript-driver)

It adds a hook to the [Session.run](https://github.com/neo4j/neo4j-javascript-driver/blob/4.3/src/session.js#L97) method and collects useful data.

Some insights about the schema:

From **beforeHook**
    
`connectionParameters.mode`: `READ` or `WRITE`
`connectionParameters.host`: db host name
`connectionParameters.port`: port
`connectionParameters.database`: database name
`connectionParameters.user`: username for database connection
`query`: the input query passed
`params`: parameters for the query

From **afterHook**
`response`: The result of the query
`summary`: Somme statistics about the query execution returned by neo4j
OR
`error`: The error thrown by the driver

Example of generated span (truncated for clarity and privacy)

```json
{
    // truncated
    "type": "neo4j",
    "started": 1608809335218,
    "connectionParameters": {
        "mode": "READ",
        "host": "db-1234567890.graphenedb.com",
        "port": 7687,
        "database": "",
        "user": "neo4j"
    },
    "query": "MATCH (u:User { id: $id }) RETURN u;",
    "params": "{\"id\":\"49e9878f-18cc-4dd4-9aa1-c6a39b5d4c12\"}",
    "response": "[{\"keys\":\"****\",\"length\":1,\"_fields\":[{\"name\":\"John\"[TRUNCATED]",
    "summary": "{\"query\":{\"text\":\"MATCH (u:User { id: $id }) RETURN u;\",\"parameters\":{\"id\":\"49e9878f-18cc-4dd4-9aa1-c6a39b5d4c12\"}},\"queryType\":\"r\",[TRUNCATED]",
    "ended": 1608809335333
}
```

### DIFF
--- a/src/hooks/neo4j.js
+++ b/src/hooks/neo4j.js
@@ -1,0 +1,73 @@
+import { safeRequire } from '../utils/requireUtils';
+import * as logger from '../logger';
+import { hook } from '../extender';
+import { getRandomId } from '../utils';
+import { SpansContainer, TracerGlobals } from '../globals';
+import { extendNeo4jSpan, createNeo4jSpan } from '../spans/neo4jSpan';
+import { payloadStringify } from '../utils/payloadStringify';
+import { getCurrentTransactionId, NEO4J_SPAN } from '../spans/awsSpan';
+
+function queryBeforeHook(args, extenderContext) {
+  const awsRequestId = TracerGlobals.getHandlerInputs().context.awsRequestId;
+  const transactionId = getCurrentTransactionId();
+  const query = args[0];
+  const params = args[1];
+  const spanId = getRandomId();
+  const started = Date.now();
+
+  const connectionHolder = this._connectionHolderWithMode(this._mode);
+  const connectionParameters = {
+    mode: this._mode,
+    host: connectionHolder._connectionProvider._seedRouter._host,
+    port: connectionHolder._connectionProvider._seedRouter._port,
+    database: connectionHolder._database,
+    user: connectionHolder._connectionProvider._authToken.principal,
+  };
+
+  const span = createNeo4jSpan(
+    transactionId,
+    awsRequestId,
+    spanId,
+    { started },
+    { connectionParameters, query, params },
+    NEO4J_SPAN
+  );
+
+  SpansContainer.addSpan(span);
+  extenderContext.currentSpan = span;
+}
+
+const createResultHook = currentSpan => originalResult => {
+  const ended = Date.now();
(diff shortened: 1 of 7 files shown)


src/hooks/neo4j.js
```diff
@@ -0,0 +1,73 @@
...
+  const extendedSpan = extendNeo4jSpan(currentSpan, {
+    ended,
+    error: payloadStringify(error),
+  });
+  SpansContainer.addSpan(extendedSpan);
+};
+
+function queryAfterHook(args, originalFnResult, extenderContext) {
+  const { currentSpan } = extenderContext;
+  originalFnResult.then(createResultHook(currentSpan)).catch(createErrorHook(currentSpan));
```
doriaviram, line 61: Just note that In this way, you adding the user `catch` handler to the promise also if he didn't ask for it.
It can change the user behavior, We didn't figure out yet what is the best way to hook promises, with the minimum side effect so currently we used `hookPromise` that effect only if the user adds  `then` and `catch` by himself, but for now, you can keep this behavior only for `neo4j` until we will decide what is the best way to instrument promises

bboure: @doriaviram Ho, that is a good point! Didn't think about that.
Just thinking out loud here, but wouldn't it work to just re-throw the Error after catching it?

Here is a quick runkit demo: https://runkit.com/bboure/wrap-promises

or am I missing something?

doriaviram: Yes, it's a good point, actually, You can use that, please add tests for all the cases you can think of, see that the exception is really thrown for example

doriaviram: It also is better implantation of our `hookPromise` in the `extender` lib

bboure: @doriaviram After some investigation and testing, it turns out that this should be working as is.

The reason for that is that we return [the original promise](https://github.com/lumigo-io/lumigo-node/blob/master/src/extender.js#L21) to the user. The `then().catch()` inside the hook has **no effect** at all on it. The original will still throw an error if `then()` (or await) is used without `catch`.
This would not be the case if we had chained them, though.

See demo: https://runkit.com/bboure/chained-promise-vs-non-chained

I also tried using `hookPromise()` first. It turns out that it does not (always) work.
When I used the `await` style promise handling, the span would not be captured.

doriaviram: Cool, I will run our integration-test and will keep you updated

PR #199, opened 2020-12-27
RD-4253 - fix: auto instrument handler with absolute path

saartochner:


auto-instrument-handler/index.js
```diff
@@ -53,7 +57,7 @@ const handler = (event, context, callback) => {
   return lumigo.trace(module[functionName])(event, context, callback);
 };
 
-module.exports = { ORIGINAL_HANDLER_KEY, handler };
+module.exports = { ORIGINAL_HANDLER_KEY, parseOriginalHandler, handler };
```
doriaviram, line 60: Put this function into some _utils: {parseOriginalHandler: () => {}}

doriaviram: Try to add here also the function that removes ourselves from the stacktrace

doriaviram (changes requested):

doriaviram (approved):

PR #211, opened 2021-01-03
feat: add wrapper script to lumigo-auto-instrument

saartochner:


auto-instrument-handler/lumigo_wrapper
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
doriaviram, line 1: rename to `auto-instrument-handler/lumigo_wrapper.sh` ?

doriaviram (approved):

auto-instrument-handler/lumigo_wrapper
saartochner: I prefer it to be without the `.sh` because that's how they wrote their examples (all our special extensions scripts looks like that) 🤔

PR #223, opened 2021-01-27
fix: better handling with wrong instrumentation [RD-4419]

doriaviram:


src/tracer.test.js
```diff
@@ -344,6 +344,20 @@ describe('tracer', () => {
...
+  test('trace; no context', async () => {
+    const token = 'DEADBEEF';
+    const lumigoTracer = require('./index')({ token });
+
+    const userHandler1 = async event => {
+      return 'ok';
+    };
+    const event = { a: 'b', c: 'd' };
+
+    await lumigoTracer.trace(userHandler1)(event);
```
nirLumigo, line 356: Don't you want to check the return value as well?

nirLumigo (approved):

src/tracer.test.js
doriaviram: I'm just want to check the wrapping here

PR #262, opened 2021-04-25
fix: start build with typescript instead of webpack

doriaviram:
* There are no code changes (only configuration), all the changes caused by the prettier version update.
* Changes only in files:
* `.babelrc.js`, `.npmignore`, `package.json`, `tsconfig.json`, `webpack.config.js`


tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
```
nirLumigo, line 1: 🔥

nirLumigo (approved):

PR #268, opened 2021-05-06
feat: add timer to track tracer added latency

OrrLumigo:


src/extender.js
```diff
@@ -16,7 +17,9 @@ export const hook = (module, funcName, options = {}, shimmerLib = shimmer) => {
       if (isFunctionAlreadyWrapped(originalFn)) return originalFn;
       return function (...args) {
         safeBeforeHook.call(this, args, extenderContext);
+        Timer.pause();
         const originalFnResult = originalFn.apply(this, args);
+        Timer.start();
```
doriaviram, line 22: If we using the decorators, this one should be removed?

src/globals.ts
```diff
@@ -1,12 +1,123 @@
 import * as logger from './logger';
 import { Context } from 'aws-lambda';
-
 const MAX_TAGS = 50;
 const MAX_TAG_KEY_LEN = 50;
 const MAX_TAG_VALUE_LEN = 50;
 const ADD_TAG_ERROR_MSG_PREFIX = 'Skipping addExecutionTag: Unable to add tag';
 export const DEFAULT_MAX_SIZE_FOR_REQUEST = 1000 * 1000;
 
+export const Timer = (() => {
```
doriaviram, line 9: Extract it to different file

doriaviram, line 9: Try to find a better name, its not just a timer that anyone can use
```suggestion
export const RuntimeTimer | 
export const GlobalTimer |
export const ... |
```

src/globals.ts
```diff
@@ -1,12 +1,123 @@
...
+  let time = 0;
+
+  type CockState = 'stopped' | 'idle' | 'counting';
+
+  let state: CockState = 'stopped';
+  let startTime = 0;
+  let _duration = 0;
+
+  // since utils uses globals had to duplicate function
```
doriaviram, line 18: Anyway lets extract it to new file

src/globals.ts
```diff
@@ -1,12 +1,123 @@
...
+  const init = (duration) => {
+    time = 0;
+    startTime = 0;
+    state = 'idle';
+    _duration = duration;
+  };
+
+  const start = () => {
+    if (state === 'stopped') {
+      init(process.env.LUMIGO_TRACER_TIMEOUT || 500);
```
doriaviram, line 43: Let's do also some calculation with the lambda timeout if it's 1s. let's define the max to be `timeout / 5 (200 ms)` max of `500ms`

src/globals.ts
```diff
@@ -1,12 +1,123 @@
...
 const MAX_TAGS = 50;
 const MAX_TAG_KEY_LEN = 50;
 const MAX_TAG_VALUE_LEN = 50;
 const ADD_TAG_ERROR_MSG_PREFIX = 'Skipping addExecutionTag: Unable to add tag';
 export const DEFAULT_MAX_SIZE_FOR_REQUEST = 1000 * 1000;
 
+export const Timer = (() => {
+  let time = 0;
+
+  type CockState = 'stopped' | 'idle' | 'counting';
```
doriaviram, line 12: Its look like you never used the `stopped` only idle and counting

src/globals.ts
```diff
@@ -1,12 +1,123 @@
...
+      time += new Date().getTime() - startTime;
+      state = 'idle';
+    }
+  };
+
+  const isTimePassed = () => {
+    let res: boolean;
+    if (state === 'counting') {
+      let now = new Date().getTime();
+      res = time + (now - startTime) > _duration;
```
doriaviram, line 62: This logic is repeated over and over again, try to reuse that

src/globals.ts
```diff
@@ -1,12 +1,123 @@
...
+        start();
+        const result = originalMethod.apply(this, args);
+        pause();
+        return result;
+      };
+
+      return descriptor;
+    };
+  };
+  return { timedSync, timedAsync, stop, getTime, getState, isTimePassed, pause, start, init };
```
doriaviram, line 118: Thinking out load: its too many functions:
Who needs to use `getTime, getState`? 
and why `start` and `stop` instead of `start` `stop` and `pause`

src/hooks/Http.ts
```diff
@@ -0,0 +1,338 @@
+import {
```
doriaviram, line 1: Why the file name start with Uppercase ?

src/hooks/Http.ts
```diff
@@ -0,0 +1,338 @@
...
+  addHeaders,
```
doriaviram, line 2: Please highlight in comments what really changed  here

src/tracer.js
```diff
@@ -34,6 +34,8 @@ export const MAX_ELEMENTS_IN_EXTRA = 10;
 export const LEAK_MESSAGE =
   'Execution leak detected. More information is available in: https://docs.lumigo.io/docs/execution-leak-detected';
 
+Timer.init(process.env.LUMIGO_TRACER_TIMEOUT || 500);
```
doriaviram, line 37: This should be called before the start each invocation, if you will leave it here its will not reset every invocation (This called now once per process [Container])

src/tracer.js
```diff
@@ -176,6 +179,7 @@ export const trace = ({ token, debug, edgeHost, switchOff, eventFilter, stepFunc
   userHandler
 ) => async (event, context, callback) => {
   try {
+    Timer.start();
```
doriaviram, line 182: Where is the stop ?

src/globals.test.js
```diff
@@ -1,8 +1,63 @@
 import * as globals from './globals';
 import { ConsoleWritesForTesting } from '../testUtils/consoleMocker';
-import { DEFAULT_MAX_SIZE_FOR_REQUEST } from './globals';
+import { DEFAULT_MAX_SIZE_FOR_REQUEST, Timer } from './globals';
 
 describe('globals', () => {
+  describe('Timer - tests', () => {
```
doriaviram, line 6: Missing tests:
 - Taking value from env vars
 - "Stopping" instrumentation after time is passed
 - Timer is reseted every invocation

jest.config.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   collectCoverage: true,
   collectCoverageFrom: [
     '**/src/**/**/**/**.js',
+    '**/src/**/**/**/**.ts',
```
doriaviram, line 8: 💪

doriaviram (changes requested):
GJ!! 💪

src/extender.js
orr-levinger: we can use then only on class methods

src/globals.ts
orr-levinger: but its kind of a global utility no?
it was on a separate file but than i thought that it fits the global

src/tracer.js
orr-levinger: it doesnt starts the timer just init it to the wanted duration

PR #271, opened 2021-05-12
fix: fix all security vulnerabilities

doriaviram:
Fixing **All** security vulnerabilities
image


src/tracer.js
```diff
@@ -172,52 +172,53 @@ export const performStepFunctionLogic = (handlerReturnValue) => {
   );
 };
 
-export const trace = ({ token, debug, edgeHost, switchOff, eventFilter, stepFunction }) => (
```
doriaviram, line 175: Its only update of `prettier`

src/parsers/aws.test.js
```diff
@@ -199,8 +199,7 @@ describe('aws parser', () => {
     const response = {
       statusCode: 200,
       receivedTime: 1564495048705,
-      body:
```
doriaviram, line 202: Its only update of `prettier`

PR #273, opened 2021-05-25
feat: add programmatic error without token

OrrLumigo:


src/index.js
```diff
@@ -45,4 +45,9 @@ module.exports = function ({
   };
 };
 
-Object.assign(module.exports, { addExecutionTag: ExecutionTags.addTag });
+Object.assign(module.exports, {
+  addExecutionTag: ExecutionTags.addTag,
+  info: LumigoLogger.info,
+  warn: LumigoLogger.warn,
+  error: LumigoLogger.error,
+});
```
doriaviram, line 53: Any tests?

doriaviram: I didnt get it yet, if `info` and `warning` do nothing. don't expose them

doriaviram: I dont think error its clear enough 
```suggestion
Object.assign(module.exports, {
  reportError: 
});
```

doriaviram (changes requested):

PR #274, opened 2021-05-26
RD-5371 auto instrument load handlers just as aws

saartochner:
Instead of parse the handler's string ourselves, I copied the parsing code from AWS official js lambda API: 
https://github.com/aws/aws-lambda-nodejs-runtime-interface-client

### DIFF
--- a/auto-instrument-handler/aws/Errors.js
+++ b/auto-instrument-handler/aws/Errors.js
@@ -1,0 +1,112 @@
+/**
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * This code was copied from:
+ * https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/main/src/Errors/index.ts
+ *
+ * Defines custom error types throwable by the runtime.
+ */
+
+"use strict";
+
+const util = require("util");
+
+function _isError(obj) {
+  return (
+    obj &&
+    obj.name &&
+    obj.message &&
+    obj.stack &&
+    typeof obj.name === "string" &&
+    typeof obj.message === "string" &&
+    typeof obj.stack === "string"
+  );
+}
+
+/**
+ * Attempt to convert an object into a response object.
+ * This method accounts for failures when serializing the error object.
+ */
+function toRapidResponse(error) {
+  try {
+    if (util.types.isNativeError(error) || _isError(error)) {
+      return {
+        errorType: error.name,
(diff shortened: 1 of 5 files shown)


auto-instrument-handler/index.test.js
```diff
@@ -3,47 +3,38 @@ const index = require('./index');
 describe('tracer', () => {
   test('handler call the original function', () => {
     process.env[index.LUMIGO_SWITCH_OFF] = 'TRUE';
+    process.env.LAMBDA_TASK_ROOT = '/var/task';
 
     process.env[index.ORIGINAL_HANDLER_KEY] = '';
     expect(() => index.handler({}, {})).toThrowError(
-      'Could not load the original handler. Are you sure that the handler is correct?'
+      'Bad handler'
```
orr-levinger, line 10: why change the original message?

auto-instrument-handler/aws/aws-user-function.js
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/**
+ * Split the handler string into two pieces: the module name and the path to
+ * the handler function.
+ */
+function _splitHandlerString(handler) {
+  let match = handler.match(FUNCTION_EXPR);
+  if (!match || match.length != 3) {
+    throw new MalformedHandlerName("Bad handler");
```
orr-levinger, line 46: That not an informative message..

auto-instrument-handler/index.test.js
saartochner: On the contrary - we used to change the original message.
Not we're raising the exact same message as the vanilla runtime :)

auto-instrument-handler/aws/aws-user-function.js
saartochner: That's not my code 😂 
This is the message that the node runtime throws in this case. I don't want that the message will be different with our instrumentation 🤔

OrrLumigo (approved):

auto-instrument-handler/aws/Errors.js
```diff
@@ -0,0 +1,112 @@
+/**
```
saartochner, line 1: This file is a copied file

auto-instrument-handler/aws/aws-user-function.js
```diff
@@ -0,0 +1,157 @@
+/**
```
saartochner, line 1: This file is a copied file

uri-p (approved):

PR #279, opened 2021-06-14
fix: fixing start span logic [RD-5495]

doriaviram:


package.json
```diff
@@ -26,22 +26,22 @@
     "shimmer": "1.2.1"
   },
   "devDependencies": {
-    "@babel/cli": "^7.14.3",
```
doriaviram, line 29: This is security updates

OrrLumigo (approved):

PR #285, opened 2021-06-21
fix: removeDuplicate implementation

saartochner:
building the previous implementation was created wrong code:
```
var __spreadArray = (this && this.__spreadArray) || function (to, from) {
    for (var i = 0, il = from.length, j = to.length; i < il; i++, j++)
        to[j] = from[i];
    return to;
};
var removeDuplicates = function (arr) { return __spreadArray([], new Set(arr)); };
```

I replaced this code will and will explore it further later.
The attached IT shows that this functionality didn't work before the change and currently works.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -507,4 +507,4 @@ export const isEmptyString = (str): boolean =>
   !!(!str || (typeof str === 'string' && str.length === 0));
 
 // @ts-ignore
-export const removeDuplicates = (arr) => [...new Set(arr)];
+export const removeDuplicates = (arr) => Array.from(new Set(arr));


saartochner:
Update: this is the bug:
https://stackoverflow.com/questions/33464504/using-spread-syntax-and-new-set-with-typescript/33464709

src/utils.ts
```diff
@@ -507,4 +507,4 @@ export const isEmptyString = (str): boolean =>
   !!(!str || (typeof str === 'string' && str.length === 0));
 
 // @ts-ignore
-export const removeDuplicates = (arr) => [...new Set(arr)];
```
doriaviram, line 510: Add a test that proves the bug fix

doriaviram (changes requested):

doriaviram (approved):

PR #292, opened 2021-08-01
fix: allow non-string pruning

saartochner:


src/utils/payloadStringify.test.js
```diff
@@ -116,6 +116,14 @@ describe('payloadStringify', () => {
     expect(result).toEqual('[2,3,3,4,5,6,7,11]...[too long]');
   });
 
+  test('prune on non-string', () => {
```
OrrLumigo, line 119: would add undefined and null tests

OrrLumigo (changes requested):

src/utils/payloadStringify.js
```diff
@@ -30,7 +31,14 @@ export const keyToOmitRegexes = () => {
   return regexesList.map((x) => new RegExp(x, 'i'));
 };
 
-export const prune = (str, maxLength) => (str || '').substr(0, maxLength);
+export const prune = (str, maxLength) => {
```
doriaviram, line 34:
```suggestion
?
```

doriaviram (changes requested):

src/utils/payloadStringify.js
saartochner: didn't get this comment 🤔

OrrLumigo (approved):

doriaviram (approved):
Please find the root cause :)

PR #296, opened 2021-08-13
fix: make the request T/O configurable

efimk-lu:
I also increased the default T/O from 250 to 300, WDYT?

### DIFF
--- a/src/httpSpansAgent.js
+++ b/src/httpSpansAgent.js
@@ -11,8 +11,11 @@ import {
 } from './utils';
 import axios from 'axios';
 
-const REQUEST_TIMEOUT = 250;
-
+const REQUEST_TIMEOUT_FLAG_MS = 'LUMIGO_REQUEST_TIMEOUT_MS'; 
+const getRequestTimeout = () => {
+  if (process.env[REQUEST_TIMEOUT_FLAG_MS]) return parseInt(process.env[REQUEST_TIMEOUT_FLAG_MS]);
+  return 300;
+};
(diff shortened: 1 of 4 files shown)


efimk-lu:
@doriaviram any specific way I should test it or the IT are enough?

src/httpSpansAgent.js
```diff
@@ -11,8 +11,11 @@ import {
 } from './utils';
 import axios from 'axios';
 
-const REQUEST_TIMEOUT = 250;
-
+const REQUEST_TIMEOUT_FLAG_MS = 'LUMIGO_REQUEST_TIMEOUT_MS';
+const getRequestTimeout = () => {
+  if (process.env[REQUEST_TIMEOUT_FLAG_MS]) return parseInt(process.env[REQUEST_TIMEOUT_FLAG_MS]);
+  return 300;
+};
```
doriaviram, line 18: Nice, move to `utils.ts`

src/httpSpansAgent.test.js
```diff
@@ -45,7 +45,7 @@ describe('HttpSpansAgent', () => {
       Authorization: 't_xxx',
       'User-Agent': '@lumigo/tracerMock$1.2.3',
     });
-    expect(requests[0].timeout).toEqual(250);
+    expect(requests[0].timeout).toEqual(300);
```
doriaviram, line 48: Add a test that its configured by env var

doriaviram (approved):

PR #300, opened 2021-09-01
feat: filter lumigo from stack trace

OrrLumigo:


src/utils.ts
```diff
@@ -376,7 +376,7 @@ export const removeLumigoFromStacktrace = (handleReturnValue) => {
     const { stack } = err;
     const stackArr = stack.split('\n');
 
-    const pattern = '/dist/lumigo.js:';
+    const pattern = 'lumigo';
```
saartochner, line 379: hmmm this is a bit problematic.
people tend to have test functions, so you see people with file name "lumigo_test", or handlers `lumigo_handler`.
Why do you change it?

efimk-lu: Agree, can you explain how it relates to the issue?

OrrLumigo: The use throws an Object and we try to filter only `dist/lumigo.js`
but the user is has a tracer that is not packed to a single file thus the stack contains a different string pattern: 

`Error: [object Object]
    at Object.normalizeLambdaError (/opt/nodejs/node_modules/@lumigo/tracer/dist/tracer.js:210:15)
    at /opt/nodejs/node_modules/@lumigo/tracer/dist/tracer.js:304:54
    at step (/opt/nodejs/node_modules/@lumigo/tracer/dist/tracer.js:33:23)
    at Object.next (/opt/nodejs/node_modules/@lumigo/tracer/dist/tracer.js:14:53)
    at fulfilled (/opt/nodejs/node_modules/@lumigo/tracer/dist/tracer.js:5:58)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)`
    
    we should include this case aswell

src/utils.ts
```diff
@@ -376,8 +380,8 @@ export const removeLumigoFromStacktrace = (handleReturnValue) => {
     const { stack } = err;
     const stackArr = stack.split('\n');
 
-    const pattern = '/dist/lumigo.js:';
-    const cleanedStack = stackArr.filter((v) => !v.includes(pattern));
+    const patterns = ['/dist/lumigo.js:', '/node_modules/@lumigo/tracer/'];
```
saartochner, line 383: change to const
add test 👌

saartochner (changes requested):
great work

src/utils.ts
```diff
@@ -366,6 +366,10 @@ export const isAwsService = (host, responseData = undefined): boolean => {
   );
 };
 
+const checkInput = (input, words) => {
```
saartochner, line 369: find a better name 😂

src/utils.ts
```diff
@@ -366,6 +366,10 @@ export const isAwsService = (host, responseData = undefined): boolean => {
...
+  return words.some((word) => new RegExp(word, 'i').test(input));
```
saartochner, line 370: define the words as regexes from the first place. You don't need to compile it into regexes every time...

src/utils.ts
```diff
@@ -366,6 +370,10 @@ export const isAwsService = (host, responseData = undefined): boolean => {
   );
 };
 
+const isLumigoStackTrace = (input) => {
+  return LUMIGO_STACK_PATTERNS.some((word) => new RegExp(word, 'i').test(input));
```
saartochner, line 374:
```suggestion
  return LUMIGO_STACK_PATTERNS.some((word) => word.test(input));
```

saartochner (approved):

PR #324, opened 2021-10-28
RD-6312 - Missing execution tags 

OrrLumigo:
fix: cannot read property 'headers' of undefined


src/reporter.test.ts
```diff
@@ -589,6 +589,40 @@ describe('reporter', () => {
     expect(reporter.forgeAndScrubRequestBody([], 100)).toEqual(undefined);
   });
 
+  test('scrubSpans missing http fields', () => {
+    const spans = [
+      {
```
saartochner, line 594: add comment, e.g. "not response"

PR #330, opened 2021-11-23
fix: fixed axios vulnerabilities

OrrLumigo:


package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "agentkeepalive": "^4.1.4",
-    "axios": "0.21.1",
+    "axios": "^0.24.0",
```
efimk-lu, line 25: Why not upgrade to 0.21.2?

OrrLumigo: i ran "npm audit fix" that suppose to fix vulnerabilities and thats what it outputed but changed it now manually

src/httpSpansAgent.test.js
```diff
@@ -38,7 +38,7 @@ describe('HttpSpansAgent', () => {
 
     const requests = AxiosMocker.getAxiosMocker().history.post;
     expect(requests.length).toEqual(1);
-    expect(requests[0].data).toEqual(reqBody);
+    expect(JSON.parse(requests[0].data)).toEqual(reqBody);
```
efimk-lu, line 41: @OrrLumigo how will affect the data arriving to our BE?

OrrLumigo: this is just the Axios mocker its not the actual Axios lib

doriaviram (approved):

PR #331, opened 2021-11-29
fix: min timeout should be DEFAULT_CONNECTION_TIMEOUT + 50

OrrLumigo:


src/utils.test.js
```diff
@@ -490,10 +490,10 @@ describe('utils', () => {
     expect(utils.getTracerMaxDurationTimeout()).toEqual(500);
   });
 
-  test('getTracerMaxDurationTimeout -> 20% of the run time', () => {
+  test('getTracerMaxDurationTimeout -> max(20%,350) of the run time', () => {
     const inputs = new HandlerInputesBuilder().withTimeout(1000).build();
     TracerGlobals.setHandlerInputs(inputs);
-    expect(utils.getTracerMaxDurationTimeout()).toEqual(200);
+    expect(utils.getTracerMaxDurationTimeout()).toEqual(350);
```
doriaviram, line 496: Change the timeout of the function to check the 20% logic

src/utils.ts
```diff
@@ -199,8 +199,7 @@ export const getTracerMaxDurationTimeout = (): number => {
   if (process.env[TRACER_TIMEOUT_FLAG]) return parseFloat(process.env[TRACER_TIMEOUT_FLAG]);
   const { context } = TracerGlobals.getHandlerInputs();
   if (isAwsContext(context)) {
-    const { remainingTimeInMillis } = getContextInfo(context);
-    return Math.min(DEFAULT_TRACER_MAX_DURATION_TIMEOUT, remainingTimeInMillis / 5);
```
doriaviram, line 203: Its looks like you deleted this logic, what about small timeout functions?

doriaviram (changes requested):

src/globals.ts
```diff
@@ -118,6 +118,9 @@ export const TracerGlobals = (() => {
     event: {},
     context: {},
   };
+
+  let lambdaTimeout = DEFAULT_TRACER_TIMEOUT;
```
saartochner, line 122: maybe "duration" is a better name?

OrrLumigo: left or right side ?

src/globals.ts
```diff
@@ -118,16 +119,30 @@ export const TracerGlobals = (() => {
...
     isStepFunction: false,
     maxSizeForRequest: DEFAULT_MAX_SIZE_FOR_REQUEST,
+    lambdaTimeout: MAX_TRACER_ADDED_DURATION_ALLOWED,
+  };
+
+  const setHandlerInputs = ({ event, context }) => {
+    Object.assign(tracerInputs, {
+      lambdaTimeout: context?.getRemainingTimeInMillis
+        ? context.getRemainingTimeInMillis()
+        : MAX_TRACER_ADDED_DURATION_ALLOWED,
```
doriaviram, line 137: leave this as undefined, let the user choose how to handle

PR #339, opened 2021-12-26
fix: when missing _X_AMZN_TRACE_ID skip tracer

OrrLumigo:


src/utils.ts
```diff
@@ -256,7 +256,8 @@ export const isSwitchedOff = (): boolean =>
     () =>
       validateEnvVar(SWITCH_OFF_FLAG) ||
       TracerGlobals.getTracerInputs().switchOff ||
-      !isValidAlias()
+      !isValidAlias() ||
+      !process.env['_X_AMZN_TRACE_ID']
```
saartochner, line 260: change to `isAwsEnvironment`

src/tracer.ts
```diff
@@ -224,12 +224,16 @@ export const trace =
       return performPromisifyType(err, data, type, callback);
     }
     context.__wrappedByLumigo = true;
+    let pStartTrace = Promise.resolve();
+    let functionSpan;
+    if (!isSwitchedOff()) {
```
saartochner, line 229: I think that this `if` should be the first line in this function (`trace`), and it should immediately stop everything (all the code lines in here should not run!)

src/tracer.test.js
```diff
@@ -482,6 +482,23 @@ describe('tracer', () => {
     expect(Http.hookHttp).toHaveBeenCalledTimes(1);
   });
 
+  test('trace; async local lambda no aws _X_AMZN_TRACE_ID', async () => {
+    delete process.env['_X_AMZN_TRACE_ID'];
```
saartochner, line 486: This line is not mandatory (because of the next line), right?

src/tracer.test.js
```diff
@@ -482,6 +482,23 @@ describe('tracer', () => {
     expect(Http.hookHttp).toHaveBeenCalledTimes(1);
   });
 
+  test('trace; async local lambda no aws _X_AMZN_TRACE_ID', async () => {
```
saartochner, line 485:
```suggestion
  test('trace; async local lambda no aws environment', async () => {
```

src/tracer.ts
```diff
@@ -197,6 +197,13 @@ export const trace =
   ({ token, debug, edgeHost, switchOff, stepFunction }) =>
   (userHandler) =>
   async (event, context, callback) => {
+    if (!isAwsEnvironment()) {
+      let handlerReturnValue = await promisifyUserHandler(userHandler, event, context);
+      handlerReturnValue = normalizeLambdaError(handlerReturnValue);
+      const cleanedHandlerReturnValue = removeLumigoFromStacktrace(handlerReturnValue);
+      const { err, data, type } = cleanedHandlerReturnValue;
+      return performPromisifyType(err, data, type, callback);
```
saartochner, line 205: why not simply:
`return userHandler(event, context)`
?

src/utils.ts
```diff
@@ -169,7 +169,12 @@ const LUMIGO_STACK_PATTERNS = [
 const validateEnvVar = (envVar: string, value: string = 'TRUE'): boolean =>
   !!(process.env[envVar] && process.env[envVar].toUpperCase() === value.toUpperCase());
 
-export const isAwsEnvironment = () => !!process.env['LAMBDA_RUNTIME_DIR'];
+export const isAwsEnvironment = () =>
+  !!(
+    process.env['LAMBDA_RUNTIME_DIR'] &&
+    !process.env['AWS_SAM_LOCAL'] && // local SAM
+    !process.env['IS_LOCAL']
+  ); // local SLS
```
saartochner, line 177:
```suggestion
    !process.env['IS_LOCAL'] // local SLS
  ); 
```

src/tracer.test.js
OrrLumigo: No i want see that its not crushing when its missing

src/tracer.ts
OrrLumigo: what if it comes with callback?

src/utils.ts
OrrLumigo: lint auto doing it

src/tracer.ts
OrrLumigo: and if there is exception we need to remove lumigo from there

saartochner: of course, add the callback.
I don't want to run any logic here. Even not the remove-from-stacktrace IMO

src/tracer.test.js
nirLumigo: Maybe we can do `parametrize`?  One with "_X_AMZN_TRACE_ID" and one without "_X_AMZN_TRACE_ID"?

nirLumigo (approved):

saartochner (approved):

PR #347, opened 2022-01-24
feat: add execution tags by env vars

saartochner:


OrrLumigo (approved):
COOL FEATURE!!

src/tracer.test.js
```diff
@@ -463,6 +464,22 @@ describe('tracer', () => {
...
+    expect(spies.setSwitchOff).not.toHaveBeenCalled();
+
+    const { context } = new HandlerInputesBuilder().build();
+
+    const userHandler4 = async (event, context, callback) => {
+      return {};
+    };
+    await lumigoTracer.trace(userHandler4)({ key1: 'aaa' }, context, jest.fn());
+
+    expect(spies.autoTagEvent).toBeCalledWith({ key1: 'aaa' });
```
doriaviram, line 480: Don't use the mock, query the ExecutionTags API and check if there is a tag

src/globals.ts
```diff
@@ -111,7 +111,11 @@ export const ExecutionTags = (() => {
   // @ts-ignore
   const clear = () => (global.tags = []);
 
-  return { addTag, getTags, clear, validateTag };
+  const autoTagEvent = (event) => {
+    getAutoTagKeys().forEach((key) => event[key] && addTag(key, event[key]));
```
doriaviram, line 115: Make sure that the `value` is string

src/tracer.test.js
saartochner: Unfortunately this is hard :(

doriaviram (approved):

PR #348, opened 2022-01-25
fix: resource name is url and not name

OrrLumigo:


src/hooks/http.ts
```diff
@@ -60,13 +66,20 @@ export class Http {
     currentSpan,
     maxSize = getEventEntitySize(true)
   ) {
+    let serviceData: AwsServiceData = {};
     if (body && !requestData.truncated) {
+      serviceData = getAwsServiceData({ ...requestData, body }, null);
```
saartochner, line 71: You better do it in the row below.
You need to use `requestData.body + body` and not only `body`.
Please add test on it

saartochner (changes requested):
GJ!

doriaviram (approved):

saartochner (approved):

PR #350, opened 2022-01-28
feat(events/getTriggeredBy): handle empty SQS event

naorpeled:
I've noticed a case where I filtered all non actionable records (s3:TestEvent in my case)
from an S3 bucket notification SQS event and received an error that the trigger source could not be found within the records.

I added code that handles it and also updated the test suite


doriaviram (approved):
Nice!

OrrLumigo (approved):
Nice

PR #383, opened 2022-06-01
RD-8226 - update README.md and layers with node16 support

nirbon:


layers/LAYERS16x.md
```diff
@@ -0,0 +1,23 @@
+Layers runtime: nodejs16.x
+----
+| Region | ARN |
+| --- | --- |
```
doriaviram, line 4: Did you create it manually? You need to find where we creating this file

doriaviram (changes requested):

layers/LAYERS16x.md
nirbon: correct, sorry i've removed it (I saw that node12 is identical to node 14 and assumed) 

lumigo-node/scripts/prepare_layer_files.sh

i think i found how to do it in the circleci builds.

doriaviram (approved):

GuyMoses (approved):

PR #387, opened 2022-06-14
fix: isSqsEvent was missing array validation

OrrLumigo:


src/parsers/eventParser.ts
```diff
@@ -82,7 +82,7 @@ export const isSnsEvent = (event): event is SNSEvent => {
 };
 
 export const isSqsEvent = (event): event is SQSEvent => {
-  return event?.Records[0]?.eventSource === 'aws:sqs';
+  return event?.Records?.[0]?.eventSource === 'aws:sqs';
```
saartochner, line 85: add test

saartochner (changes requested):
GJ! Add a test please

saartochner (approved):

src/parsers/eventParser.test.ts
```diff
@@ -411,6 +411,12 @@ describe('event parser', () => {
     );
   });
 
+  test('unknownEvent parse', () => {
+    const unknownEvent = { a: 'a' };
+    const parsedEvent = parseEvent(unknownEvent);
+    expect(unknownEvent).toEqual(parsedEvent);
+  });
```
saartochner, line 418: This test doesn't check your change...

saartochner (changes requested):
GJ!

src/parsers/eventParser.test.ts
OrrLumigo: you are right !!

src/parsers/eventChecker.ts
```diff
@@ -0,0 +1,33 @@
+import {
```
doriaviram, line 1: All tests should be in `filename.test.ts`

doriaviram (changes requested):

src/parsers/eventParser.test.ts
```diff
@@ -411,6 +412,23 @@ describe('event parser', () => {
...
+      jest.spyOn(EventChecker, 'isSqsEvent'),
+      jest.spyOn(EventChecker, 'isS3Event'),
+      jest.spyOn(EventChecker, 'isCloudfrontEvent'),
+    ];
+    EventParser.parseEvent(unknownEvent);
+    spies.forEach((spy) => {
+      expect(spy.mock.results[0].type).toEqual('return');
+      expect(spy.mock.results[0].value).toEqual(false);
+      spy.mockRestore();
+    });
```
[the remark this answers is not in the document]
OrrLumigo: we need the spies because we call it from eventParser and the catch is in eventParser

src/parsers/eventChecker.ts
OrrLumigo: I extracted to the chekcer because spies don't work on inner exported calls

saartochner (approved):

PR #389, opened 2022-07-03
feat: support lambda -> sqs -> sns -> lambda

saartochner:


.circleci/config.yml
```diff
@@ -78,6 +78,9 @@ jobs:
           cd ..
           git clone git@github.com:lumigo-io/larn.git
     - checkout
+    - run: apt-get update
+    - run: apt-get install jq
+    - run: npm install -g aws-cdk@1.125.0
```
efimk-lu, line 83: @moshe-shaham-lumigo why do you need CDK?

moshe-shaham-lumigo: because all integration tests now work with the new edge

efimk-lu: So all services will need this as part of their circle config?

moshe-shaham-lumigo: yes
everyone that uses the orb already have it

efimk-lu: ok, I would recommend updating our base docker at https://github.com/lumigo-io/utils/blob/master/deployment/ci-docker/Dockerfile it will improve our CI running time

moshe-shaham-lumigo (approved):

PR #394, opened 2022-09-07
fix: parse ddb partial request body

saartochner:


src/utils.ts
```diff
@@ -195,6 +195,14 @@ export const safeGet = (obj, arr, dflt = null) => {
   return current || dflt;
 };
 
+export const safeJsonParse = (obj, dflt = undefined) => {
```
doriaviram, line 198:
```suggestion
export const safeJsonParse = (obj, default = undefined) => {
```

doriaviram: You can use `safeExecute` inside

doriaviram (approved):

src/utils.ts
saartochner: Moved to `safeExecute`,
But couldn't change to `default` because this is a reserved word

PR #398, opened 2022-09-28
feat: in W3C header injection: add option to request even if does'nt exists

saartochner:


src/hooks/http.ts
```diff
@@ -123,6 +123,29 @@ export class Http {
...
 
+  static addOptionsToHttpRequestArguments(originalArgs, newOptions) {
+    // We're switching on the different signatures of http:
+    // https://nodejs.org/api/http.html#httpgeturl-options-callback
+    if (typeof originalArgs[0] === 'string' || originalArgs[0] instanceof URL) {
+      if (originalArgs[1]) {
+        if (typeof originalArgs[1] === 'function') {
+          // The signature is: (url, callback)
+          originalArgs.push(originalArgs[1]);
+          originalArgs[1] = newOptions;
```
nirLumigo, line 134: For each option, can you also add a comment for the new signature?

src/hooks/http.ts
```diff
@@ -191,11 +214,14 @@ export class Http {
       if (isRequestToAwsService && !isKeepHeadersOn()) {
         const { awsXAmznTraceId } = getAWSEnvironment();
         const traceId = getPatchedTraceId(awsXAmznTraceId);
-        options.headers['X-Amzn-Trace-Id'] = traceId;
+        headers && (headers['X-Amzn-Trace-Id'] = traceId);
```
nirLumigo, line 217: Was this change needed? there is a default value for headers in line 205

src/hooks/http.test.js
```diff
@@ -838,6 +875,49 @@ describe('http hook', () => {
     expect(span.info.httpInfo.request.headers[TRACESTATE_HEADER_NAME]).toBeDefined();
   });
 
+  test('wrapHttpLib - check request - headers exists', () => {
```
nirLumigo, line 878: What is the purpose of the test? Verifying that `TRACESTATE_HEADER_NAME` exists? Please add it in the description of the test. Same regarding the next test.

nirLumigo (approved):
Nice! :)
Added small comments, please check before merging.

src/hooks/http.ts
saartochner: Not use options is the important change here :)

PR #399, opened 2022-10-15
feat: MongoDB 4.x support

mmanciop:
* Added support for MongoDB 4.x 
* Improved extender.js to be able to wrap functions that are constructors (which is needed by MongoDB to ensure that the `monitorCommands` flag is set in the client on instantiation 

TODOs:
* Tests + Itests
* Support for detecting failures and timeouts on connection


mmanciop:
In reality is probably better to pass an option to hook in case we wrap a constructor and invoke the wrapped function precisely once, as we do not know whether there are side effects. After all, we should always know whether the wrapped function is a constructor or not.

mmanciop (commented):
I have seen with ClariFruit's Luke that our instrumentation breaks MongoClient.connect

PR #401, opened 2022-10-23
fix: disable adding execution tags when lambda is not traced

vladlLumigo:


src/globals.ts
```diff
@@ -95,10 +101,15 @@ export const ExecutionTags = (() => {
 
   const addTag = (key, value, shouldLogErrors = true) => {
     try {
-      logger.debug(`Adding tag: ${key} - ${value}`);
-      if (!validateTag(key, value, shouldLogErrors)) return false;
-      // @ts-ignore
-      global.tags.push({ key: normalizeTag(key), value: normalizeTag(value) });
+      if (isLambdaTraced()) {
+        logger.debug(`Adding tag: ${key} - ${value}`);
+        if (!validateTag(key, value, shouldLogErrors)) return false;
```
therightstuff, line 106: please use braces (`{`, `}`) even for single lines

src/globals.ts
```diff
@@ -95,10 +101,15 @@ export const ExecutionTags = (() => {
...
+        // @ts-ignore
```
therightstuff, line 107: why do we need this `@ts-ignore`? can we put the reason inline?

src/globals.ts
```diff
@@ -95,10 +101,15 @@ export const ExecutionTags = (() => {
...
+        global.tags.push({ key: normalizeTag(key), value: normalizeTag(value) });
+      } else {
+        shouldLogErrors && logger.warnClient(`${ADD_TAG_ERROR_MSG_PREFIX}: lambda is not traced`);
+        return false;
+      }
     } catch (err) {
       shouldLogErrors && logger.warnClient(ADD_TAG_ERROR_MSG_PREFIX);
```
therightstuff, line 114: do you know why the error isn't included in the client warning?

src/utils.ts
```diff
@@ -586,3 +586,7 @@ export const filterObjectKeys = (
     .reduce((cur, key) => {
       return Object.assign(cur, { [key]: obj[key] });
     }, {});
+
+export const isKillSwitchOn = () => !!process.env['LUMIGO_SWITCH_OFF'];
```
therightstuff, line 590: please verify that this is correct, if i recall correctly this should be a string comparison with very specific cases where it's "true"

src/utils.ts
```diff
@@ -586,3 +586,7 @@ export const filterObjectKeys = (
...
+
+export const isLambdaTraced = () => !!(!isKillSwitchOn() && isAwsEnvironment());
```
therightstuff, line 592: both of these methods should return a boolean value, so we should not need `!!`. for increased readability, please put the positive check first (ie. `=> (isAwsEnvironment() && !isKillSwitchOn())`)

therightstuff (changes requested):
i do have a few concerns i've highlighted, but it's mostly great :)

src/utils.ts
doriaviram, line 590: There is a method for that `isSwitchedOff`

src/globals.ts
vladlLumigo: Good question, it was in the original code, I don't know what was the reason

src/globals.test.js
```diff
@@ -445,4 +453,26 @@ describe('globals', () => {
...
+  test.each`
+    killSwitchValue | isAwsEnvironment | expectedRetValue | expectedTags
+    ${false}        | ${true}          | ${true}          | ${[{ key: 'k0', value: 'v0' }]}
+    ${true}         | ${true}          | ${false}         | ${[]}
+    ${true}         | ${false}         | ${false}         | ${[]}
+    ${false}        | ${false}         | ${false}         | ${[]}
+  `(
+    'killSwitch=$killSwitchValue, isAwsEnv=$isAwsEnvironment, expectedRetValue=$expectedRetValue$',
+    ({ killSwitchValue, isAwsEnvironment, expectedRetValue, expectedTags }) => {
+      const oldEnv = Object.assign({}, process.env);
```
doriaviram, line 466: You don't need to save the envs. the infra do it for you.
Look at steup.js

src/globals.test.js
```diff
@@ -445,4 +453,26 @@ describe('globals', () => {
 
     process.env = { ...oldEnv };
   });
+
+  test.each`
```
doriaviram, line 457: It looks like its hard to maintain these spaces, maybe array its better

doriaviram (changes requested):

src/globals.ts
vladlLumigo: done

src/utils.ts
vladlLumigo: used already existing function isSwitchedOff() instead

src/utils.ts
vladlLumigo: done

src/utils.ts
vladlLumigo: done

src/globals.test.js
vladlLumigo: done

src/globals.test.js
vladlLumigo: tried both versions,  decided to stay with the table version

therightstuff (approved):

src/globals.ts
vladlLumigo: we show the client only specific errors, a general catch , will contain information that the client don't really need to see

doriaviram (approved):

doriaviram (approved):

PR #403, opened 2022-11-08
fix: increase clarity when reporting unhandled promise rejections

therightstuff:


therightstuff:
Further investigation has led to further questions, I'm closing this PR until we have more clarity ourselves.

PR #407, opened 2022-11-18
fix: mysql-2: support extract values in all known calling conventions

mmanciop:
Collect the values passed as named or unnamed parameters in to `mysql-2`'s `.query()` and `.execute()` for the following calling conventions:

* Values nested in `options` object: `client.query({ sql: 'SELECT * FROM users WHERE id = ?', values: ['123']}, callback?)`
* Single value: `client.query('SELECT * FROM users WHERE id = ?', '123', callback?)`
* Named parameters: `client.query('SELECT * FROM users WHERE id = ?id', {id: '123'}, callback?)`

### DIFF
--- a/src/hooks/mySql.js
+++ b/src/hooks/mySql.js
@@ -36,6 +36,7 @@ function extractValuesFromArg(arg) {
   if (isObject(arg) && typeof arg === 'object' && !!arg['values']) {
     return arg['values'];
   }
+  return undefined;
 }
 
 function queryBeforeHook(args, extenderContext) {
--- a/src/hooks/mySql.js
+++ b/src/hooks/mySql.js
@@ -32,11 +32,22 @@ function extractQueryFromArg(arg) {
   }
 }
 
-function extractValuesFromArg(arg) {
+function extractValuesFromQueryOptions(arg) {
   if (isObject(arg) && typeof arg === 'object' && !!arg['values']) {
     return arg['values'];
   }
   return undefined;
+}
+
+function extractValuesFromQueryParameter(arg) {
+  return (
+    // This is for the `.query(sql, single_param, callback?)` form; we show it as a JSON array for simplicity
+    (typeof arg === 'string' ? arg : undefined) ||
+    // This is for named parameters in the `.query(sql, {p1:'foo', p2:'bar'}, callback?)` form
+    (isObject(arg) && typeof arg === 'object' ? arg : undefined) ||
+    // This is for unnamed parameters in the `.query(sql, ['foo', 'bar'], callback?)` form
+    (Array.isArray(arg) ? arg : [])
+  );
 }
 
 function queryBeforeHook(args, extenderContext) {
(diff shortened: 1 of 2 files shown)


doriaviram (approved):

src/hooks/mySql.js
```diff
@@ -32,11 +32,26 @@ function extractQueryFromArg(arg) {
   }
 }
 
+function extractValuesFromArg(arg) {
+  if (isObject(arg) && typeof arg === 'object' && !!arg['values']) {
+    return arg['values'];
+  }
+}
```
therightstuff, line 39: please add an explicit return, even if it's null/undefined

src/hooks/mySql.js
```diff
@@ -32,11 +32,26 @@ function extractQueryFromArg(arg) {
...
+
 function queryBeforeHook(args, extenderContext) {
   const awsRequestId = TracerGlobals.getHandlerInputs().context.awsRequestId;
   const transactionId = getCurrentTransactionId();
   const query = extractQueryFromArg(args[0]);
-  const values = Array.isArray(args[1]) ? args[1] : [];
+
+  const values =
+    // This is for the `.query(options, callback?)` form
+    extractValuesFromArg(args[0]) ||
```
therightstuff, line 48: i would argue that this method should be named `extractValuesFromQueryOptions` and that the second "half" of the ternary be extracted to a method named `extractValuesFromQueryParameter`

src/hooks/mySql.test.js
```diff
@@ -52,279 +52,338 @@ describe('mySql', () => {
...
-      expect(spans).toEqual([
-        createBaseBuilderFromSpan(spans[0])
-          .withQuery('Unknown')
-          .withConnectionParameters(DUMMY_OPTIONS)
-          .withResponse(createExpectedResponse())
-          .build(),
-      ]);
-      done();
+    test('hook -> query (text: string, callback: Function) -> fail', (done) => {
+      const error = new Error('DuumyError');
```
therightstuff, line 90: trivial typo: DummyError (copy/pasted 12 times)

therightstuff (changes requested):
looking good! one minor request and one trivial observation, but overall happiness

src/hooks/mySql.js
mmanciop: Fixed in 5fbb948901e88348c95947aa27909090528b9fde

src/hooks/mySql.js
mmanciop: Fixed in dc7a11fbf24d18f3fa4ca6f207ead57ec93b0910

src/hooks/mySql.test.js
mmanciop: Fixed in b1086a53a98675ac5ccae59c5c31b5303d8d404d

therightstuff (approved):
excellent! thank you for the fixes :)

Issue #431, opened 2023-03-13
Tracer should handle traceIdArr greater than 3 and should not be throwing unhandled errors


builtbylane:
https://github.com/lumigo-io/lumigo-node/blob/c4e3228043030170956f899f5ce28c924b183999/src/utils.ts#L79

```typescript;
export const getTraceId = (awsXAmznTraceId) => {
  if (!awsXAmznTraceId) {
    throw new Error('Missing _X_AMZN_TRACE_ID in Lambda Env Vars.');
  }

  const traceIdArr = awsXAmznTraceId.split(';');
  if (traceIdArr.length !== 3) {
    throw new Error('Expected 3 semi-colon separated parts in _X_AMZN_TRACE_ID.');
  }

  const traceId = {};
  // XXX Populates Root, Parent and Sampled keys.
  traceIdArr.forEach((item) => {
    const [key, value] = item.split('=');
    traceId[key] = value;
  });

  if (!traceId['Root'] || !traceId['Parent'] || !traceId['Sampled']) {
    throw new Error(`Either Root, Parent or Sampled weren't found in traceId.`);
  }

  // @ts-ignore
  const transactionId = traceId.Root.split('-')[2];

  // @ts-ignore
  traceId.transactionId = transactionId;

  return traceId;
};
```

Can this be written without ts-ignores?
what does the // XXX comment mean?
Why check that the traceIdArr is exactly 3, if those values are being checked lower down?

Note: this code will break systems that every pass in a `awsXAmznTraceId` that has extra information attached.

builtbylane:
Without a full refactor, this would make the code safer. 
```
  if (traceIdArr.length >= 3) {
    throw new Error('Expected at least 3 semi-colon separated parts in _X_AMZN_TRACE_ID.');
  }
```

doriaviram:
Hey, @builtbylane Thanks for reporting, we are checking the issue

builtbylane:
this is a redacted but real _X_AMZN_TRACE_ID that breaks lumigo: `Root=1-640f97b4-12345;Parent=1234;Sampled=0;Lineage=1234:0`

as a first step, the code should not strictly be looking for exactly 3 parts. 

As a followup, this code feels unsafe for production systems. There are too many areas where an unexpectedly formed traceId could blow up a lambda. For instance, what if the Root had fewer than 2 hyphens? Is there a reason an error like that should block the execution of a lambda?

Nxtra:
[...]

PR #432, opened 2023-03-13
fix: lineage in _x_amzn_trace_id

saartochner:


doriaviram:
fixes https://github.com/lumigo-io/lumigo-node/issues/431

PR #434, opened 2023-03-15
RD-10354 - feat: provide finer granularity to secret masking

saartochner:


testUtils/httpSpanBuilder.js
```diff
@@ -1,5 +1,7 @@
 import { lowerCaseObjectKeys } from '../src/utils';
 import { payloadStringify } from '../src/utils/payloadStringify';
+import {isString} from "@lumigo/node-core/lib/utils";
+import {json} from "caniuse-lite/data/features";
```
GuyMoses, line 4: where is it used?

testUtils/httpSpanBuilder.js
```diff
@@ -130,15 +132,17 @@ export class HttpSpanBuilder {
     this._span.info.httpInfo.host = host;
 
     const headers = HttpSpanBuilder.addHeader(this._span.info.httpInfo.request.headers, { host });
-    this._span.info.httpInfo.request.headers = headers;
+    this._span.info.httpInfo.request.headers = JSON.stringify(headers);
```
GuyMoses, line 135: so do we change the property types?

src/utils.ts
```diff
@@ -621,3 +633,13 @@ export const filterObjectKeys = (
...
+export const getRequestBodyMaskingRegex = (): string | undefined =>
+  process.env[LUMIGO_SECRET_MASKING_REGEX_HTTP_REQUEST_BODIES];
+export const getRequestHeadersMaskingRegex = (): string | undefined =>
+  process.env[LUMIGO_SECRET_MASKING_REGEX_HTTP_REQUEST_HEADERS];
+export const getResponseBodyMaskingRegex = (): string | undefined =>
+  process.env[LUMIGO_SECRET_MASKING_REGEX_HTTP_RESPONSE_BODIES];
+export const getResponseHeadersMaskingRegex = (): string | undefined =>
+  process.env[LUMIGO_SECRET_MASKING_REGEX_HTTP_RESPONSE_HEADERS];
+export const getEnvVarsMaskingRegex = (): string | undefined =>
+  process.env[LUMIGO_SECRET_MASKING_REGEX_ENVIRONMENT];
```
GuyMoses, line 645: add this to the README documentation?

testUtils/httpSpanBuilder.js
saartochner: yes - the BE can handle both.
The problem with the existing `payloadStringify` method is that it doesn't "remember" paths, so it is hard to refer to specific paths inside a big payload - we have to do it here

saartochner: I'm taking it back. I'll upload a fix the doesn't do it.

GuyMoses (approved):

PR #435, opened 2023-03-22
fix: replace auto-instrument.handler with NODE_OPTIONS injection

mmanciop:
The `auto-instrument.handler` has some issues related with complex CommonJS cases on Node.js 16+, and it is very hard to cover all the possible combinations of how Node.js modules of the function handler should be loaded [^1][^2] without breaking others.

Instead, this PR skips the entire auto-instrument.handler finicky thing, and introduces a simple NODE_OPTIONS based method inside the `AWS_LAMBDA_EXEC_WRAPPER`. The method is effectively the same we use with the Lumigo OpenTelemetry Distro for JS.

[^1] CommonJS: https://nodejs.org/api/modules.html
[^2] ESM: https://nodejs.org/api/esm.html

### DIFF
--- a/auto-instrument-handler/lumigo_wrapper
+++ b/auto-instrument-handler/lumigo_wrapper
@@ -6,13 +6,13 @@ shopt -s nocasematch
 
 function log {
     if [ "${LUMIGO_DEBUG}" = 'true' ]; then
-        echo "$1"
+        echo "#LUMIGO# $1"
     fi
 }
 
 # LUMIGO_NODE_OPTIONS_INJECTION has default true
 if [ "${LUMIGO_NODE_OPTIONS_INJECTION:-true}" = 'true' ]; then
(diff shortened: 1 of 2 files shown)


auto-instrument-handler/lumigo_wrapper
```diff
@@ -1,5 +1,40 @@
 #!/bin/bash
 
-export LUMIGO_ORIGINAL_HANDLER=$_HANDLER
-export _HANDLER="lumigo-auto-instrument.handler"
+# Enable case-insentive match of vars
```
therightstuff, line 3:
```suggestion
# Enable case-insensitive match of vars
```

mmanciop:
Since this also requires an internal extension to be able to collect the event and the function context, we decided to do something else in the short run, namely #436.

PR #438, opened 2023-04-20
RD-10627-no-execution-tags-added-when-lambda-times-out

shanishiri:


src/tracer/tracer.ts
```diff
@@ -246,8 +248,19 @@ const setupTimeoutTimer = () => {
...
+        const transactionId = getCurrentTransactionId();
+        const awsRequestId: string = context.awsRequestId;
+
+        const enrichmentSpan = generateEnrichmentSpan(
+          ExecutionTags.getTags(),
+          token,
+          transactionId,
+          awsRequestId
+        );
+        const spansToSend = [...spans, enrichmentSpan];
```
saartochner, line 261: Small performance enhance: you don't need to copy the array, just `push` to it

src/tracer/tracer.test.js
```diff
@@ -109,6 +110,36 @@ describe('tracer', () => {
...
+    TracerGlobals.setHandlerInputs({ event, context });
+    const functionSpan = getFunctionSpan(event, context);
+    tracer.startTrace(functionSpan).then(() => {
+      const span1 = { id: 'SomeRandomHttpSpan' };
+      SpansContainer.addSpan(span1);
+      globals.ExecutionTags.addTag('key1', 'value1');
+
+      setTimeout(() => {
+        const requests = AxiosMocker.getRequests();
+        //1 for start span, 1 for SomeRandomHttpSpan + 1 enrichmentSpan
```
saartochner, line 131: the comment is a bit misleading. I'm not sure that I understood it myself...
Maybe change it to: "2 requests: 1 for the start span, and another for the timeout spans" 🤔

saartochner (approved):
LGTM

PR #444, opened 2023-05-22
feat: use improved @lumigo/node-core secret scrubbing

mmanciop:


src/reporter.ts
```diff
@@ -89,23 +95,33 @@ const scrubSpan = (span) => {
...
+          ScrubContext.HTTP_RESPONSE_BODY,
           sizeLimit,
           span.info.httpInfo.response.truncated
         );
       }
       if (span.info.httpInfo.request?.body) {
         span.info.httpInfo.request.body = scrub(
           decodeHttpBody(request.body, isError),
-          request.headers,
+          response.headers,
```
kenfinnigan, line 106: Is this intended to scrub the response headers and not the request as it did previously?

src/hooks/http.test.js
```diff
@@ -1270,9 +1270,6 @@ describe('http hook', () => {
     expect(Http.scrubQueryParams('?password=123&plain=123&secret=123')).toEqual(
       '?password=****&plain=123&secret=****'
     );
-    process.env[LUMIGO_SECRET_MASKING_REGEX_HTTP_QUERY_PARAMS] = '["plain"]';
```
kenfinnigan, line 1273: Is this test no longer required?

kenfinnigan (commented):
Couple of comments, likely for my understanding more than anything

README.md
```diff
@@ -78,8 +78,13 @@ You can read more about it [here](https://www.typescriptlang.org/tsconfig#esModu
...
-  * We support more granular masking using the following parameters. If not given, the above configuration is the fallback: `LUMIGO_SECRET_MASKING_REGEX_HTTP_REQUEST_BODIES`, `LUMIGO_SECRET_MASKING_REGEX_HTTP_REQUEST_HEADERS`, `LUMIGO_SECRET_MASKING_REGEX_HTTP_RESPONSE_BODIES`, `LUMIGO_SECRET_MASKING_REGEX_HTTP_RESPONSE_HEADERS`, `LUMIGO_SECRET_MASKING_REGEX_HTTP_QUERY_PARAMS`.
+* `LUMIGO_SECRET_MASKING_REGEX='["regex1", "regex2"]'`: Prevents Lumigo from sending keys that match the supplied regular expressions in process environment data, HTTP headers, payloads and queries. All regular expressions are case-insensitive. The "magic" value `all` will redact everything. By default, Lumigo applies the following regular expressions: `[".*pass.*", ".*key.*", ".*secret.*", ".*credential.*", ".*passphrase.*"]`. More fine-grained settings can be applied via the following environment variables, which will override `LUMIGO_SECRET_MASKING_REGEX` for a specific type of data:
```
kenfinnigan, line 81: This is tied to a question I had on https://github.com/lumigo-io/opentelemetry-java-distro/issues/21.

Saying "Prevents Lumigo from sending keys" is not reflected with your comment that it's still sent but masked. We should clarify the language to prevent any confusion over what's actually happening

mmanciop: Ah now I understand

src/reporter.ts
mmanciop: That's a big, fat bug you just found

src/hooks/http.test.js
mmanciop: It's redundant. That is already tested extensively in node-core, and it would take an overhaul of the tests using jest.isolateModules to test effectively because of the way node-core works.

src/reporter.ts
mmanciop: The impact actually was that, if the response did not have a JSON-like MIME type, but the request was, we might fail to scrub the latter. Thanks for catching it.

src/reporter.test.ts
```diff
@@ -318,7 +318,7 @@ describe('reporter', () => {
               },
               response: {
                 truncated: true,
-                body: `{"a":"${value}","b":"${value}","key":"****","password":"****","e":"${value}","secret":"****","f":"${value}","g":"${value}"}...[too long]`,
+                body: `{"a":"${value}","b":"${value}","key":"****","password":"****","e":"${value}","secret":"****","f":"${value}","g":"${value}"}✂`,
```
therightstuff, line 321: i'm not sure how i feel about this, because it is cute - aside from it being less clear and difficult to identify, are we sure there's no chance of it causing problems in the frontend?

src/reporter.test.ts
```diff
@@ -362,6 +361,7 @@ describe('reporter', () => {
         },
         { dummyEnd },
       ];
+      const shorter = 'a'.repeat(1944);
```
therightstuff, line 364: do we not have a constant for the max length?

src/reporter.test.ts
```diff
@@ -375,7 +375,7 @@ describe('reporter', () => {
               },
               response: {
                 truncated: false,
-                body: `{"key":"****","password":"****","e":"${value}","secret":"****","f":"${value}","g":"${value}","h":"${shorter}"}...[too long]`,
+                body: `{"key":"****","password":"****","e":"${value}","secret":"****","f":"${value}","g":"${value}","h":"${shorter}✂"}`,
```
therightstuff, line 378: regardless of what truncation indicator we go with, i would suggest using a constant to avoid brittleness

src/reporter.test.ts
```diff
@@ -638,10 +640,42 @@ describe('reporter', () => {
...
+                statusCode: 200,
+                receivedTime,
+              },
+            },
+          },
+        },
+        { dummyEnd },
+      ];
+
+      expect(spans[0]?.info?.httpInfo.request.body).toBe('{"secret":"fo');
```
therightstuff, line 676: should the "fo" not be masked in this case?

src/reporter.ts
```diff
@@ -53,22 +53,35 @@ export const shouldTrim = (spans, maxSendBytes: number): boolean => {
...
   } catch (e) {
-    return payloadStringify(payload, sizeLimit, truncated);
+    logger.warn('An error occurred while stringifying JSON payload', e);
   }
+
+  return payloadStringify(payload, scrubContext, sizeLimit, null, truncated);
 }
 
+/*
+ * TODO! We should _not_ scrub spans *again*, this already done (or should be done) in the
```
therightstuff, line 79: please remove the TODOs and create tickets to do the work instead

src/spans/awsSpan.ts
```diff
@@ -328,14 +318,16 @@ export const getHttpSpan = (
 
   const prioritizedSpanId = getHttpSpanId(randomRequestId, spanId);
   if (requestData) {
-    requestData.body && (requestData.body = shallowMask('requestBody', requestData.body));
+    requestData.body &&
```
therightstuff, line 321: i would appreciate using full if conditions for readability

src/spans/sqlSpan.ts
```diff
@@ -24,13 +24,13 @@ export const createSqlSpan = (
...
-      user: dbFields.connectionParameters.user,
+      host: dbFields?.connectionParameters?.host,
+      port: dbFields?.connectionParameters?.port,
+      database: dbFields?.connectionParameters?.database,
+      user: dbFields?.connectionParameters?.user,
     },
-    query: normalizeQuery(dbFields.query),
-    values: dbFields.values ? payloadStringify(dbFields.values) : '',
+    query: normalizeQuery(dbFields?.query),
+    values: payloadStringify(dbFields?.values) || '',
```
therightstuff, line 33: wouldn't `payloadStringify(dbFields?.values)` always result in a value?

therightstuff (approved):
mostly looking good, nothing blocking but i do have a few concerns that i'd like addressed

src/reporter.test.ts
mmanciop: It has been tested and it's already shipped with the OpenTelemetry JS distro

src/reporter.test.ts
mmanciop: Yes but it is maddening to calculate backwards what the size oughta be in the test. In reality, the red flag is if the leftover characters change in an unrelated PR.

src/reporter.test.ts
mmanciop: Actually, I am very much against using constants in tests: you DO want to know if something breaks the contract with the backend. That character is the contract. If you can change it with just one keystroke, it's way too easy to break the contract.

src/reporter.test.ts
```diff
@@ -440,14 +440,16 @@ describe('reporter', () => {
...
-      const actual = JSON.parse(reporter.forgeAndScrubRequestBody(spans, expectedResultSize));
+      const actual = JSON.parse(reporter.forgeAndScrubRequestBody(spans, expectedResultSize)!);
       expect(actual).toEqual(expected);
     });
 
     test('forgeAndScrubRequestBody long request', () => {
       const value = 'a'.repeat(10);
       const long = 'a'.repeat(getEventEntitySize(true));
-      const shorter = 'a'.repeat(getEventEntitySize());
+      const shorter = 'a'.repeat(
```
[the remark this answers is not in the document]
mmanciop: Same as https://github.com/lumigo-io/lumigo-node/pull/444#discussion_r1203890803

src/reporter.test.ts
mmanciop: No, because that thing is not actually JSON. If we untruncate it, we lie in the payload for the likely reason of issues to the end user.

src/reporter.ts
mmanciop: I am against this. JIRA is where important items like these go to die. I want each and every one of us looking at this code go "DAMN we should do that" :D @dori-admin wdyt.

src/spans/awsSpan.test.js
```diff
@@ -247,9 +237,14 @@ describe('awsSpan', () => {
...
-    process.env[LUMIGO_SECRET_MASKING_REGEX_ENVIRONMENT] = '["DONT_TELL"]';
-    process.env.DONT_TELL = '1234';
-    expect(JSON.parse(awsSpan.getEnvsForSpan()).DONT_TELL).toEqual('****');
+    jest.isolateModules(() => {
+      process.env['LUMIGO_SECRET_MASKING_REGEX_ENVIRONMENT'] = '["DONT_TELL"]';
+      process.env['DONT_TELL'] = '1234';
+
+      const awsSpan = require('./awsSpan');
+
+      expect(JSON.parse(awsSpan.getEnvsForSpan()).DONT_TELL).toEqual('****');
```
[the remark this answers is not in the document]
mmanciop: > the mask always four characters

This. See https://github.com/lumigo-io/tracing-ingestion.

src/spans/awsSpan.ts
mmanciop: There is no logical difference with the code that was already merged, but sure, I'll send a patch

src/spans/sqlSpan.ts
mmanciop: Fair

kenfinnigan (approved):

PR #457, opened 2023-08-06
RD-10988-extend-secret-scrubbing-and-automatic-tag-extraction-in-the-lambda-node-js-tracer-to-be-able-to-handle-nested-stringified-json

shanishiri:


saartochner:
Hi, why not on `node-common`?
https://github.com/lumigo-io/node-core/blob/master/src/common/secrets.ts
Also, I would suggest that you do it behind an env var. This could be a significant performance issue.

src/globals.ts
```diff
@@ -77,6 +78,11 @@ export const ExecutionTags = (() => {
   // @ts-ignore
   global.tags = [];
 
+  interface AutoTagEvent {
+    event: string;
+    keyToEventMap: any;
```
nadav3396, line 83: This should be map, no?

src/globals.ts
```diff
@@ -134,10 +140,35 @@ export const ExecutionTags = (() => {
...
+  const getValue = (autoTagEvent: AutoTagEvent, innerKey) => {
+    let obj = autoTagEvent.event;
+    const eventByKey = autoTagEvent.keyToEventMap;
+    if (eventByKey && eventByKey[innerKey]) {
+      obj = eventByKey[innerKey];
+      return obj && { event: obj, keyToEventMap: eventByKey };
+    }
+    try {
+      if (obj && isString(obj) && obj[innerKey] === undefined) {
+        const parsedObj = JSON.parse(obj);
```
nadav3396, line 152: You don't reuse parsed keys in your map.

src/globals.ts
```diff
@@ -134,10 +140,35 @@ export const ExecutionTags = (() => {
...
+        eventByKey[innerKey] = parsedObj[innerKey];
+        return parsedObj && { event: parsedObj[innerKey], keyToEventMap: eventByKey };
+      }
+    } catch (err) {}
+    if (obj && obj[innerKey]) {
```
nadav3396, line 157: I think this condition should be before the try.

src/globals.ts
```diff
@@ -134,10 +140,35 @@ export const ExecutionTags = (() => {
...
+      eventByKey[innerKey] = obj[innerKey];
+      return { event: obj[innerKey], keyToEventMap: eventByKey };
+    }
+    return { event: undefined, keyToEventMap: eventByKey };
```
nadav3396, line 161: Maybe its better to add a condition to handle when we got undefined

src/globals.ts
```diff
@@ -134,10 +140,35 @@ export const ExecutionTags = (() => {
...
+
   const autoTagEvent = (event) => {
+    let keyToEventMap: {} = {};
     getAutoTagKeys().forEach((key) => {
-      const value = key.split('.').reduce((obj, innerKey) => obj && obj[innerKey], event);
-      value && addTag(key, value);
+      const value: AutoTagEvent = key
+        .split('.')
+        .reduce(getValue, { event: event, keyToEventMap: keyToEventMap });
+      keyToEventMap = { ...keyToEventMap, ...value.keyToEventMap };
```
nadav3396, line 170: Why ```keyToEventMap = value.keyToEventMap ;``` is not enough?

src/globals.ts
```diff
@@ -134,10 +140,35 @@ export const ExecutionTags = (() => {
...
+    let obj = autoTagEvent.event;
+    const eventByKey = autoTagEvent.keyToEventMap;
+    if (eventByKey && eventByKey[innerKey]) {
+      obj = eventByKey[innerKey];
+      return obj && { event: obj, keyToEventMap: eventByKey };
+    }
+    try {
+      if (obj && isString(obj) && obj[innerKey] === undefined) {
+        const parsedObj = JSON.parse(obj);
+        eventByKey[innerKey] = parsedObj[innerKey];
```
nadav3396, line 153: you use the inner key and not the full key to map the object and this can cause bugs for example in:
```
{
   a.duplicated_inner_key: "first_value",
   b.duplicated_inner_key: "second_value",
}
```

src/globals.ts
shanishiri: then we return undefined

src/globals.ts
shanishiri: it's an Object

src/globals.ts
```diff
@@ -134,10 +141,44 @@ export const ExecutionTags = (() => {
...
+        eventByKey[relativeKey] = parsedObj;
+        return (
+          parsedObj && {
+            event: parsedObj[innerKey],
+            keyToEvent: eventByKey,
+            relativeKey: relativeKey,
+          }
+        );
+      }
+    } catch (err) {}
```
doriaviram, line 170: Lets add debug log, that we thought its should be JSON but its not

doriaviram (approved):

PR #458, opened 2023-08-07
Rd 11126 lambda python node tracers sqs json support

sagivoululumigo:
# Features
* Added support for SQS messages sent with the new JSON protocol (see [AWS docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-json-faqs.html))

### DIFF
--- a/src/parsers/aws.ts
+++ b/src/parsers/aws.ts
@@ -1,4 +1,4 @@
-import { md5Hash, parseQueryParams, removeDuplicates, safeGet, safeJsonParse } from '../utils';
+import { md5Hash, parseQueryParams, removeDuplicates, safeGet, safeJsonParse, caseInsensitiveGet } from '../utils';
 import { traverse } from '../tools/xmlToJson';
 import * as logger from '../logger';
 import { getW3CMessageId } from '../utils/w3cUtils';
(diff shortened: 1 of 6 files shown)


nadav3396 (approved):
LGTM

src/parsers/aws.ts
```diff
@@ -122,36 +129,64 @@ export const eventBridgeParser = (requestData, responseData) => {
...
-        'MessageId',
-      ],
-      undefined
-    );
+  let awsServiceData = {};
+
+  // Note: Currently json protocol is new and not commonly used, so the default case is XML.
+  // In the future when JSON usage is default you my want to switch so JSON is the default and XML
+  // protocol is specifically checked for
+  if (
```
doriaviram, line 140: Make sure we know to extract the `resourceName` in JSON also from other SQS methods like [sendMessageBatch](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SQS.html#sendMessageBatch-property)

src/parsers/aws.ts
```diff
@@ -122,36 +129,64 @@ export const eventBridgeParser = (requestData, responseData) => {
 };
 
 export const sqsParser = (requestData, responseData) => {
-  const { body: reqBody } = requestData;
+  const { body: reqBody } = requestData || {};
```
doriaviram, line 132: Let's split this function into 2, XML & JSON, to make it more readable

src/utils.ts
```diff
@@ -268,6 +268,19 @@ export const safeGet = (obj, arr, dflt = null) => {
...
 };
+/**
+ * Finds a value from object with case-insensitive match of keys.
+ * If multiple matching keys are found, the first is returned
+ *
+ * @param obj Object to get value from
+ * @param key The key to search (can be upper / lower case or a mix of the two, it doesn't matter)
+ * @param dflt Default value if no match is found
+ */
+export const caseInsensitiveGet = (obj, key, dflt = null) => {
```
doriaviram, line 279: Add types

doriaviram (changes requested):

src/parsers/aws.ts
sagivoululumigo: I tested for this in the ```sqsParser -> happy flow JSON format```, the test is run on a single message & a batch message & the message id & resource name are tested

src/parsers/aws.ts
sagivoululumigo: Fixed

src/parsers/aws.ts
sagivoululumigo: Done, fixed it

doriaviram (approved):

PR #462, opened 2023-08-28
feat: change W3C headers propagation to be ON by default

saartochner:


src/hooks/http.test.js
```diff
@@ -904,12 +905,12 @@ describe('http hook', () => {
       .withRequestTimesFromSpan(spans[0])
       .build();
 
+    spans.forEach((span) => delete span.info.messageId);
```
doriaviram, line 908: Can we test it instead of delete?

doriaviram (approved):

PR #463, opened 2023-09-26
ci: verify pr title semantic format

sagivoululumigo:


.github/workflows/verify-pr-details.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  # Job that checks the title of the PR & makes sure it fits the semantic versioning format
+  check-pr-title:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR Title
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
```
kenfinnigan, line 23: Is there a difference between this custom script and [this](https://github.com/lumigo-io/opentelemetry-java-distro/blob/main/.github/workflows/push-actions.yml#L23)

sagivoululumigo: Good to know! switched to the action use suggested

PR #466, opened 2023-11-01
feat: very basic implementation of fetch

saartochner:


src/hooks/fetch.ts
```diff
@@ -0,0 +1,149 @@
...
+    const headersObject = {};
+    let numKeysFound = 0;
+    try {
+      // convert map to object
+      for (const [key, value] of headers) {
+        headersObject[key] = value;
+        numKeysFound++;
+      }
+    } catch (err) {
+      console.log('fetch headers as map parse error', err);
```
doriaviram, line 19: Use our logger (we not print in case we don't have env-var)

PR #469, opened 2023-12-26
feat: support Prisma 5

harelmo-lumigo:


src/hooks/prisma.test.js
```diff
@@ -0,0 +1,191 @@
...
+import { PrismaSpanBuilder } from '../../testUtils/prismaSpanBuilder';
+import { SpansContainer, TracerGlobals } from '../globals';
+import { getRandomId } from '../utils';
+import { hookPrisma } from './prisma';
+
+describe('Prisma', () => {
+  describe('bad Prisma versions', () => {
+    afterEach(() => {
+      // Reset overrides made to the prisma-client after require() in each test
+      jest.resetModules();
```
saartochner, line 11: this is actually really cool

src/hooks/prisma.test.js
```diff
@@ -0,0 +1,191 @@
...
+
+    test('produces simple operation span', async () => {
+      const count = await client.user.count({ where: { id: 123456 } });
+
+      const spans = SpansContainer.getSpans();
+      expect(spans).toHaveLength(1);
+
+      const expectedSpan = new PrismaSpanBuilder()
+        .withId(spans[0].id)
+        .withStarted(spans[0].started)
```
saartochner, line 66: can you explicitly check the this is actually filled? I.e. that `spans[0].started` is not undefined?
same for other fields

src/hooks/prisma.test.js
```diff
@@ -0,0 +1,191 @@
...
+
+      const expectedRawQuerySpan = new PrismaSpanBuilder()
+        .withId(rawQuerySpan.id)
+        .withStarted(rawQuerySpan.started)
+        .withEnded(rawQuerySpan.ended)
+        .withOperation('$queryRaw')
+        .withQueryArgs(
+          `{"values":["${userId}"],"strings":["SELECT * FROM User WHERE id = ",";"]}`
+        )
+        .withResult(`[]`)
```
saartochner, line 185: can you add a test with a very long query / result? (i.e. `select * from users`)
we want to trim long contents

src/hooks/prisma.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  prismaLib.PrismaClient = function (...args) {
+    const originalInstance = new OriginalConstructor(...args);
+
+    if (typeof originalInstance.$extends !== 'function') {
+      return originalInstance;
+    }
+
+    return originalInstance.$extends({
```
saartochner, line 65: do you want to return the response from the `$extends` function or `originalInstance`?
Can you add all your code (and specifically all the *added* code) to `safeExecute`?

src/hooks/prisma.ts
```diff
@@ -0,0 +1,71 @@
...
+      query: {
```
saartochner, line 66: Just to make sure that we're not missing something - what are other hooks that we can do (besides `query`)?

src/spans/awsSpan.ts
```diff
@@ -50,6 +50,7 @@ export const MSSQL_SPAN = 'msSql';
 export const MYSQL_SPAN = 'mySql';
 export const NEO4J_SPAN = 'neo4j';
 export const ENRICHMENT_SPAN = 'enrichment';
+export const PRISMA_SPAN = 'prisma';
```
saartochner, line 53: Don't forget: I want to see it e2e in Lumigo. Let's make sure that you'll not get "span_type not found"

src/spans/common.ts
```diff
@@ -0,0 +1,11 @@
+import { filterObjectKeys, getEventEntitySize } from '../utils';
+import { payloadStringify, truncate } from '../utils/payloadStringify';
+
+export const normalizeQuery = (query: string | any): string => {
+  if (typeof query === 'string') return truncate(query, getEventEntitySize());
```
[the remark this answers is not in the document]
saartochner: maybe I'm missing something, but I think that you don't use it in your code 😂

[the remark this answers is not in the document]
saartochner: please make sure to do it to the response as well

testUtils/prismaSetup.js
```diff
@@ -0,0 +1,8 @@
+import { execSync } from 'child_process'
+
+export default () => {
+  // Clear previous test database and migrations used by Prisma
+  execSync('rm -rf ./testUtils/prisma/migrations && rm -rf ./testUtils/prisma/dev.db')
+  // Create a new database to be used by the Prisma tests
+  execSync('npx prisma migrate dev --name init --schema ./testUtils/prisma/schema.prisma')
```
saartochner, line 7: just to make sure:`npx` is something that is always installed with the runtime? is it additional requirement? Can you maybe do it in plain node (instead of `import { execSync } from 'child_process'`).
<not mandatory>

testUtils/prismaSpanBuilder.js
```diff
@@ -0,0 +1,88 @@
...
+      },
+      memoryAllocated: '1024',
+      messageVersion: 2,
+      parentId: HttpSpanBuilder.DEFAULT_PARENT_ID,
+      reporterAwsRequestId: HttpSpanBuilder.DEFAULT_PARENT_ID,
+      readiness: 'cold',
+      region: HttpSpanBuilder.DEFAULT_REGION,
+      invokedArn: HttpSpanBuilder.DEFAULT_ARN,
+      invokedVersion: HttpSpanBuilder.DEFAULT_VERSION,
+      runtime: 'AWS_Lambda_nodejs8.10',
```
saartochner, line 25: 👨‍🦳

testUtils/prismaSpanBuilder.js
```diff
@@ -0,0 +1,88 @@
+import { PRISMA_SPAN } from '../src/spans/awsSpan';
+import { HttpSpanBuilder } from './httpSpanBuilder';
+import { payloadStringify } from '@lumigo/node-core/lib/common';
+
+export class PrismaSpanBuilder {
+  constructor() {
+    this._span = {
```
saartochner, line 7: you copied a function_span, so it contains all kinds of fields that you real http_span will not have (e.g. readiness)

saartochner (changes requested):
Very good job! 👏 
I would like to see the IT and the bottomline UI before the merge

src/hooks/prisma.ts
harelmo-lumigo: > do you want to return the response from the $extends function or originalInstance?

We have to return the output of `$extends`, as it returns a new client with the extended functionality and does not mutate `originalInstance` (actually, that's why I've added [this test](https://github.com/lumigo-io/lumigo-node/blob/ce130b9fdd9525538b8a87eecac9e720ef447aa9/src/hooks/prisma.test.js#L151)).

> Can you add all your code to safeExecute?

I guess you mean something like this:
```js
return safeExecute(
  () => originalInstance.$extends({
    query: {
      $allOperations: queryExtension,
    },
  }), 
  originalInstance
);
```
to return the original instance if `$extends` failed? yeah, it definitely makes sense

src/hooks/prisma.ts
harelmo-lumigo: All hooks are listed [here](https://www.prisma.io/docs/orm/prisma-client/client-extensions).
TL;DR - the other hooks are meant to _add functionality_ to the client / model level or mutate the query results, while `query` allows actually tapping into all operations

src/spans/awsSpan.ts
harelmo-lumigo: You mean actual a test added to the _integration test_ repo with a Lambda running Prisma, or an E2E one-time "experiment" before we launch this?

src/spans/common.ts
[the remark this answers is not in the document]
harelmo-lumigo: You're totally right, I've used it in an older iteration of the code but dropped it as it's irrelevant. Will revert the extraction to a shared module

testUtils/prismaSetup.js
harelmo-lumigo: `npx` comes bundled with `npm` since npm@5, so I don't a reason why it won't work (we use npm@6, which I need to bump up actually - as the repo is working with Node 14 that Prisma does not support. Adam got us covered in his will [PR](https://github.com/lumigo-io/utils/pull/501/) 😂 )

src/spans/awsSpan.ts
saartochner: The first is preferred, but if it gets rough then let go with the second :)

src/hooks/prisma.ts
saartochner: 👍

testUtils/prismaSetup.js
```diff
@@ -0,0 +1,8 @@
+import { execSync } from 'child_process'
+
+export default () => {
+  // Clear previous test database and migrations used by Prisma
+  execSync('rm -rf ./testUtils/prisma/migrations && rm -rf ./testUtils/prisma/dev.db')
```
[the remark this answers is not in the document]
harelmo-lumigo: Basically it's expanded to be relative to the process's CWD, which is the root of the project in that case, but I've change it to be more explicit [here](e436435af08b3e70eb3f5330d31aa4be746e3753):

testUtils/prismaSpanBuilder.js
harelmo-lumigo: Just keepin' the tradition 👴

src/hooks/prisma.test.js
harelmo-lumigo: Sure thing: 55742ae1027952ce285741cf3f65e3560d113876

testUtils/prismaSpanBuilder.js
harelmo-lumigo: Got you... I copied `MongoSpanBuilder` which had all this :( I'll remove all the irrelevant fields

src/hooks/prisma.ts
harelmo-lumigo: 74c0301bea933dfd8e95c3e551f5a1bcb6f0046a

harelmo-lumigo:
> Very good job! 👏 I would like to see the IT and the bottomline UI before the merge

Thanks :) 

Actually, I wanted to test it first in my personal dev-env. 
Do we have some kind of a formalized method to publish a "preview version" of the tracer to npm, or should I just install a locally-produced `tar.gz` on my lambda like a חורני and test it that way?

src/hooks/prisma.test.js
saartochner: also ended 🙏

testUtils/prismaSpanBuilder.js
harelmo-lumigo: Wait, I actually don't get it :) if the mongo-span has all those attributes too (e.g. `readiness`), is it by accident and those get ignored in the ingestion, or am I missing something here?

saartochner: are you sure that the mongo-span has these attributes in production as well, or is it there only in tests?
It makes no sense to have `readiness` on an HTTP span. So if it's there, then this is a bug (and then don't fix it now - just open a ticket :) )

harelmo-lumigo: The mongo-span uses [getBasicChildSpan](https://github.com/lumigo-io/lumigo-node/blob/2475be801a8db3eda04c23ce97b2125bff05a01b/src/spans/mongoDbSpan.ts#L12) which in turn uses [getBasicSpan](https://github.com/lumigo-io/lumigo-node/blob/2475be801a8db3eda04c23ce97b2125bff05a01b/src/spans/awsSpan.ts#L297), returning a bunch of attributes (such as [readiness](https://github.com/lumigo-io/lumigo-node/blob/2475be801a8db3eda04c23ce97b2125bff05a01b/src/spans/awsSpan.ts#L116)) that are probably pointless in that context, so you're definitely right :) 
I'll open a ticket for cleaning this up.

saartochner: awesome 💪

saartochner (approved):

PR #474, opened 2024-01-10
chore(deps): upgrade axios to 1.6.5

harelmo-lumigo:


package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "@lumigo/node-core": "1.13.0",
     "agentkeepalive": "^4.1.4",
-    "axios": "0.24.0",
+    "axios": "1.6.5",
```
sagivoululumigo, line 29: Are we sure we want to force customers to update to this version? it might conflict with their existing axios version.

We could set it to a base version or above, but then we have to test for every version to make sure we don't break...

harelmo-lumigo: It does not mean that client must use this version - in Node, each package can have its own version of a dependency, and is isolated from the others (contrary to Python)

sagivoululumigo: Got it, cool

sagivoululumigo (approved):
LGTM

PR #476, opened 2024-01-15
feat: dynamic errors spans payload & span prioritization

nadav3396:


src/reporter.ts
```diff
@@ -115,22 +114,61 @@ export function scrubSpans(resultSpans: any[]) {
...
+  spans.sort(spansPrioritySorter);
+  let currentSize = 0;
+  const spansToSendSizes = {};
+  const spansToSend = {};
+
+  // First we try to take only the spans metadata
+  for (let index = 0; index < spans.length; index++) {
+    const spanMetadata = getSpanMetadata(spans[index]);
+    spansToSendSizes[index] = 0;
+    if (spanMetadata == undefined) continue;
```
harelmo-lumigo, line 128: Not much of a difference in this case, but `===` is the de-facto standard:
```suggestion
    if (spanMetadata === undefined) continue;
```

src/spans/awsSpan.ts
```diff
@@ -66,6 +68,79 @@ export const getSpanInfo = (): SpanInfo => {
...
+  const span1Priority = getSpanPriority(span1);
+  const span2Priority = getSpanPriority(span2);
+
+  if (span1Priority < span2Priority) {
+    return -1;
+  } else if (span1Priority > span2Priority) {
+    return 1;
+  } else {
+    return 0;
+  }
```
harelmo-lumigo, line 94:
```suggestion
  return Math.sign(span1Priority - span2Priority)
```

src/spans/awsSpan.ts
```diff
@@ -66,6 +68,79 @@ export const getSpanInfo = (): SpanInfo => {
...
+};
+
+export const getSpanMetadata = (span: any): any => {
+  const spanCopy = JSON.parse(JSON.stringify(span));
```
harelmo-lumigo, line 98: So the benchmark is cool with that? :)

harelmo-lumigo (approved):

src/spans/awsSpan.ts
nadav3396: yes, So I think we are good. (We do this logic only in rare cases so most of the time it will not run at all

PR #481, opened 2024-05-06
feat: add big spans support (zip)

omrilumigo:
Spans in big invocations with a big payload might be scraped and dropped.
To not lose any the payload will be zipped when the LUMIGO_SUPPORT_LARGE_INVOCATIONS flag is set to true and the payload is bigger than the limit.

In addition I fixed a ts issue in the tests about type validation.

### DIFF
--- a/src/reporter.ts
+++ b/src/reporter.ts
@@ -9,12 +9,15 @@ import {
   safeExecute,
   shouldScrubDomain,
   spanHasErrors,
+  shouldTryZip,
 } from './utils';
 import * as logger from './logger';
 import { HttpSpansAgent } from './httpSpansAgent';
 import { payloadStringify } from './utils/payloadStringify';
 import { decodeHttpBody, getSpanMetadata, spansPrioritySorter } from './spans/awsSpan';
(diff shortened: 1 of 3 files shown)


src/reporter.ts
```diff
@@ -153,18 +157,33 @@ export function getPrioritizedSpans(spans: any[], maxSendBytes: number): any[] {
...
+    return undefined;
+  }
 
   if (
     (!isPruneTraceOff() && spans.length > NUMBER_OF_SPANS_IN_REPORT_OPTIMIZATION) ||
     size > maxSendBytes
   ) {
+    if (shouldTryZip) {
+      logger.debug(`Trying to zip spans, size [${size}], bigger than: [${maxRequestSize}]`);
+      const zippedSpans = gzipSync(JSON.stringify(spans)).toString('base64');
```
harelmo-lumigo, line 178: `gzipSync` blocks the main thread, might pose a performance hit and choke the lambda - can we use the async version? maybe we can zip the payload during the sending part, not when forging the body?

harelmo-lumigo (approved):

PR #489, opened 2024-07-07
fix: sqs json format no body caused exception

sagivoululumigo:


src/parsers/aws.ts
```diff
@@ -131,8 +131,8 @@ const sqsParserJsonProtocol = (requestData, responseData) => {
   const { body: resBody } = responseData || {};
   let awsServiceData = {};
 
-  const parsedReqBody = safeJsonParse(reqBody, {});
-  const parsedResBody = safeJsonParse(resBody, {});
+  const parsedReqBody = reqBody ? safeJsonParse(reqBody, {}) : undefined;
```
harelmo-lumigo, line 134: 1. I'd expect `safeJsonParse` to handle `null` / `undefined` cases - maybe we should fix it there instead of changing all usages
2. Let's add a test for it either way

sagivoululumigo: 1. It does handle that case, but it prints a stack trace error.
2. Sure, adding a test for this case

harelmo-lumigo: What I meant is that maybe this:

```js
export const safeJsonParse = (obj, dflt = undefined) => {
  return safeExecute(
    () => JSON.parse(obj),
    'Failed to parse json',
    logger.LOG_LEVELS.DEBUG,
    dflt
  )();
```
should change to:
```
export const safeJsonParse = (obj, dflt = undefined) => {
  return safeExecute(
    () => obj === undefined ? dflt : JSON.parse(obj),
    'Failed to parse json',
    logger.LOG_LEVELS.DEBUG,
    dflt
  )();
```
not a big deal, but it would spare the need to remember to check the value prior to each call to `safeJsonParse`

harelmo-lumigo (approved):

PR #490, opened 2024-07-10
fix: size and time limitations on large invocations

sagivoululumigo:


src/globals.test.js
```diff
@@ -3,6 +3,7 @@ import * as globals from './globals';
 import {
   DEFAULT_MAX_SIZE_FOR_REQUEST,
   DEFAULT_MAX_SIZE_FOR_REQUEST_ON_ERROR,
+  DEFAULT_MAX_SIZE_FOR_SPANS_SIZE_STORED_IN_MEMORY,
```
harelmo-lumigo, line 6: This represents the limit on a single span, no? so it should be `SPAN` and not `SPANS`.
Also, I think we should remove the redundant `SIZE` here and name this `DEFAULT_MAX_SIZE_FOR_SPAN_STORED_IN_MEMORY`

src/globals.ts
```diff
@@ -5,23 +5,32 @@ import { BasicSpan } from './types/spans/basicSpan';
...
 export const DEFAULT_MAX_SIZE_FOR_REQUEST_ON_ERROR = 1024 * 990;
+export const DEFAULT_MAX_SIZE_FOR_SPANS_SIZE_STORED_IN_MEMORY =
+  DEFAULT_MAX_SIZE_FOR_REQUEST_ON_ERROR * 10;
 export const MAX_TRACER_ADDED_DURATION_ALLOWED = 750;
 export const MIN_TRACER_ADDED_DURATION_ALLOWED = 200;
 
+const warnSpansSizeOnce = runOneTimeWrapper((threshold: number, currentSize: number) => {
+  logger.info(
+    `Lumigo tracer is no longer collecting data on the invocation, because it reached the maximum ` +
+      `size of calls collected of ${threshold} bytes (current size is ${currentSize} bytes)`
```
harelmo-lumigo, line 30: I would suggest:
```suggestion
    `Lumigo tracer is no longer collecting data on the invocation - maximum size of total spans collected (${threshold} bytes allowed, current size is ${currentSize} bytes)`
```
Also, maybe it's better to show KB rather than bytes?

src/globals.ts
```diff
@@ -33,12 +42,18 @@ export class SpansContainer {
       this.totalSpans += 1;
     }
     // Memory optimization, take up to 10x maxSize because of smart span selection logic
-    if (spanHasErrors(span) || getMaxRequestSizeOnError() * 10 > this.currentSpansSize) {
+    const maxSpansSize = getMaxSizeForStoredSpansInMemory();
+    if (spanHasErrors(span) || maxSpansSize > this.currentSpansSize) {
```
harelmo-lumigo, line 46: I think it's a bit clearer to reverse the condition, and also make this `<=` (in other place, we truncate the payload if it's greater than the limit, meaning a value equal to the limit is allowed - so I think we better be consistent)
```suggestion
    if (spanHasErrors(span) || this.currentSpansSize <= maxSpansSize) {
```

src/globals.ts
```diff
@@ -33,12 +42,18 @@ export class SpansContainer {
...
       this.spans[span.id] = span;
       this.currentSpansSize += getJSONBase64Size(span);
       logger.debug('Span created', span);
       return true;
     }
+
+    logger.debug('Span was not added due to size limitations', {
+      currentSpansSize: this.currentSpansSize,
```
harelmo-lumigo, line 54: Let's add the limit here as well

src/globals.ts
```diff
@@ -241,7 +257,15 @@ export const TracerGlobals = (() => {
     maxSizeForRequest = null,
     maxSizeForRequestOnError = null,
     lambdaTimeout = MAX_TRACER_ADDED_DURATION_ALLOWED,
-  }: TracerOptions) =>
+    maxSizeForStoredSpansInMemory = DEFAULT_MAX_SIZE_FOR_SPANS_SIZE_STORED_IN_MEMORY,
+  }: TracerOptions) => {
+    const parsedMaxSizeForRequestOnError =
+      maxSizeForRequestOnError ||
+      (process.env['LUMIGO_MAX_SIZE_FOR_REQUEST_ON_ERROR']
+        ? parseInt(process.env.LUMIGO_MAX_SIZE_FOR_REQUEST_ON_ERROR)
```
harelmo-lumigo, line 265: Do we (want to) handle a `NaN` case for this value?

src/globals.test.js
sagivoululumigo: This is the size limit for all the spans stored in memory. if we are already storing 10 MB of spans in memory and we want to add another then the user will be shown a message and the spans will be dropped.

Fixing the duplicate `SIZE` in the name

src/globals.ts
sagivoululumigo: I don't know if we should start manipulating the units, especially if the size limitation that the user specifies is always in bytes.

If the user could specify `10 MB` in the env var value then yes I would agree, but otherwise I think we should always refer to the size as bytes to avoid complications

src/globals.ts
sagivoululumigo: I don't want to change the logic of this env var, any case I think we should remove it when we have better support for large invocations

harelmo-lumigo (approved):

src/globals.ts
sagivoululumigo: Done

PR #493, opened 2024-07-16
feat: collect and return to fe tracer dropped spans reasons

sagivoululumigo:


src/globals.ts
```diff
@@ -29,21 +29,36 @@ const warnSpansSizeOnce = runOneTimeWrapper((threshold: number, currentSize: num
     `Lumigo tracer is no longer collecting data on the invocation - maximum size of total spans collected (${threshold} bytes allowed, current size is ${currentSize} bytes)`
   );
 }, {});
+export enum droppedSpanReasons {
```
harelmo-lumigo, line 32: Usually the JS convention is `DroppedSpanReasons`

harelmo-lumigo (approved):

src/globals.ts
sagivoululumigo: Fixed

PR #495, opened 2024-07-17
feat: report span counts as enrichment span

sagivoululumigo:


src/globals.ts
```diff
@@ -41,21 +41,25 @@ export enum DroppedSpanReasons {
...
+  drops: number;
+};
+
 export class SpansContainer {
-  private static spans: { [id: string]: BasicSpan } = {};
+  private static spans: { [id: string]: GenericSpan } = {};
   private static currentSpansSize: number = 0;
   private static totalSpans: number = 0;
-  private static droppedSpansReasons: { [reason: string]: number } = {};
+  private static droppedSpansReasons: { [reason: string]: DroppedSpansData } = {};
```
moshe-shaham-lumigo, line 52: why not just have a map of { string: number }
why is the type DroppedSpansData needed?

src/globals.ts
```diff
@@ -83,22 +87,27 @@ export class SpansContainer {
...
     numOfDroppedSpans: number = 1
   ): void {
-    this.droppedSpansReasons[reason] =
-      (this.droppedSpansReasons[reason.valueOf()] || 0) + numOfDroppedSpans;
+    // Check if the reason exists in the droppedSpansReasons object
+    // If it doesn't exist, initialize
+    if (!this.droppedSpansReasons[reason.valueOf()]) {
+      this.droppedSpansReasons[reason.valueOf()] = { drops: 0 };
+    }
+    this.droppedSpansReasons[reason.valueOf()].drops += numOfDroppedSpans;
```
moshe-shaham-lumigo, line 95: i think it's weird to use "valueOf()". why not just use the enum as the key of the map?
but anyway, not a big deal

src/reporter.ts
```diff
@@ -127,12 +138,22 @@ export function scrubSpans(resultSpans: any[]) {
   return resultSpans.filter((span) => safeExecute(scrubSpan, 'Failed to scrub span')(span));
 }
 
-export function getPrioritizedSpans(spans: any[], maxSendBytes: number): any[] {
+export function getPrioritizedSpans(
+  spans: any[],
+  maxSendBytes: number,
+  addEnrichmentSpan: boolean = true
```
moshe-shaham-lumigo, line 144: did i miss something, or this parameter is never used?

src/reporter.ts
```diff
@@ -153,32 +176,35 @@ export function getPrioritizedSpans(spans: any[], maxSendBytes: number): any[] {
...
+      SpansContainer.recordDroppedSpan(
+        DroppedSpanReasons.SPANS_SENT_SIZE_LIMIT,
+        false,
+        spansDropped
+      );
+      if (!enrichmentSpansDropped) {
+        finalSpansToSend = addOrUpdateEnrichmentSpan(finalSpansToSend);
+      } else {
+        logger.warn(
+          'Enrichment span was dropped due to size limit of total spans sent to lumigo, so dropped spans counts might be missing'
```
moshe-shaham-lumigo, line 201: don't we want to give the enrichment span priority?

src/types/spans/basicSpan.ts
```diff
@@ -12,7 +12,12 @@ export interface SpanInfo {
   logStreamName: string;
 }
 
-export interface BasicSpan {
+export interface GenericSpan {
+  id: string;
+  [key: string]: any;
+}
+
+export interface BasicSpan extends GenericSpan {
```
moshe-shaham-lumigo, line 20: why this change?

src/reporter.ts
sagivoululumigo: Your correct, I left this by mistake here.
Removed it

src/reporter.ts
sagivoululumigo: We do, but the function span has a higher priority.
We shouldn't reach this point, the only scenario when we will reach this is if the metadata of the function start + end spans + the metadata of the enrichment span is over the max send limit.

In any case I rather get the lambda invocation fact with no enrichment data then only enrichment data with no lambda invocation fact

src/globals.ts
sagivoululumigo: Your right, fixed it

moshe-shaham-lumigo (approved):

PR #500, opened 2024-08-13
fix: remove lumigo from stack when error is raised from a promise after the lambdas execution finished

eyalgolan:
**Description**: if an unhandled promise exception is raised after a lambda's execution was finished, then the logic in `removeLumigoFromStacktrace` didn't run on that exception's stacktrace. Therefore the tracer's lines were part of the stacktrace presented. 

**Jira ticket**: see [here](https://lumigo.atlassian.net/browse/RD-13400).

### DIFF
--- a/auto-instrument-handler/index.js
+++ b/auto-instrument-handler/index.js
@@ -21,19 +21,22 @@ const getHandlerAsync = async () => {
   );
 };
 
+export const removeLumigoFromError = (stacktrace) => {
+  const stackArr = stacktrace.split('\n');
+
+  const patterns = ['/dist/lumigo.js:', 'auto-instrument'];
+  const cleanedStack = stackArr.filter(v => !patterns.some(p => v.includes(p)));
+
+  return cleanedStack.join('\n');
+}
+
 const removeLumigoFromStacktrace = err => {
   // Note: this function was copied from utils.js. Keep them both up to date.
   try {
     if (!err || !err.stack) {
       return err;
     }
-    const { stack } = err;
-    const stackArr = stack.split('\n');
-
-    const patterns = ['/dist/lumigo.js:', 'auto-instrument'];
-    const cleanedStack = stackArr.filter(v => !patterns.some(p => v.includes(p)));
-
-    err.stack = cleanedStack.join('\n');
+    err.stack = removeLumigoFromError(err.stack);
 
     return err;
   } catch (e) {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,5 +1,9 @@
 import { Buffer } from 'buffer';
-import { MAX_TRACER_ADDED_DURATION_ALLOWED, MIN_TRACER_ADDED_DURATION_ALLOWED, TracerGlobals } from './globals';
+import {
+  MAX_TRACER_ADDED_DURATION_ALLOWED,
+  MIN_TRACER_ADDED_DURATION_ALLOWED,
+  TracerGlobals,
+} from './globals';
 import { isAwsContext } from './guards/awsGuards';
 import * as logger from './logger';
 import { AwsEnvironment, ContextInfo, LambdaContext } from './types/aws/awsEnvironment';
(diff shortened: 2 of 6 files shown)


src/utils.ts
```diff
@@ -480,21 +480,27 @@ const isLumigoStackTrace = (input) => {
...
-    const { stack } = err;
-    const stackArr = stack.split('\n');
 
-    const cleanedStack = stackArr.filter((v) => !isLumigoStackTrace(v));
-
-    err.stack = cleanedStack.join('\n');
+    const sanitizedErr = removeLumigoFromError(err);
 
-    return { err, data, type };
+    return { sanitizedErr, data, type };
```
harelmo-lumigo, line 503: This changes the signature of `removeLumigoFromStacktrace` from returning:
```js
{ err: err, data: data, type: type }
```
to:
```js
{ sanitizedErr: sanitizedErr, data: data, type: type }
```
so whoever deconstructs (unpacks) the value the old way:
```js
const { err, data, type } = removeLumigoFromStacktrace(...)
```
will get `err` as `undefined` - probably not what you meant (and it means that the current function is untested :( )

I think you better keep this the same by changing it to:
```suggestion
    return { err: sanitizedErr, data, type };
```

src/utils.ts
```diff
@@ -480,21 +480,27 @@ const isLumigoStackTrace = (input) => {
   return LUMIGO_STACK_PATTERNS.some((word) => word.test(input));
 };
 
+export const removeLumigoFromError = (err) => {
```
harelmo-lumigo, line 483: Thinking about this again, I think it's confusing to have a function that takes `err`, modifies it and returns the same object.
I would consider changing it to a pure function that sanitizes the stack as a `string`, and returns a new string that the caller would use to override the original value. WDYT?

src/utils.ts
eyalgolan: Oh, I didn't realize that's how it works. Thanks for explaining. I'll change this.

src/utils.ts
eyalgolan: That makes sense, no need to move around the entire `err` object. Changed, thanks!

src/utils.ts
```diff
@@ -480,19 +480,22 @@ const isLumigoStackTrace = (input) => {
...
 
+export const removeLumigoFromError = (stacktrace: string) => {
+  const stackArr = stacktrace.split('\n');
+
+  const cleanedStackArr = stackArr.filter((v) => !isLumigoStackTrace(v));
+
+  return cleanedStackArr.join('\n');
+};
 export const removeLumigoFromStacktrace = (handleReturnValue) => {
   // Note: this function was copied to the auto-instrument-handler. Keep them both up to date.
```
eyalgolan, line 491: what is the auto-instrument-handler? should I add these changes there as well? @harelmo-lumigo

auto-instrument-handler/index.js
```diff
@@ -21,19 +21,22 @@ const getHandlerAsync = async () => {
...
 
+export const removeLumigoFromError = (stacktrace) => {
+  const stackArr = stacktrace.split('\n');
+
+  const patterns = ['/dist/lumigo.js:', 'auto-instrument'];
+  const cleanedStack = stackArr.filter(v => !patterns.some(p => v.includes(p)));
+
+  return cleanedStack.join('\n');
+}
+
```
eyalgolan, line 32: Updated for now just to get the ball rolling @harelmo-lumigo

harelmo-lumigo: Cool, I see now why it has to be replicated :( (though we can change this in the future, but definitely not as a part of this PR)

harelmo-lumigo (approved):

maplesteve:
@eyalgolan, @harelmo-lumigo, @doriaviram This change breaks the layer and Lambdas crash with a RunTime error. -> https://github.com/lumigo-io/lumigo-node/issues/502

PR #513, opened 2024-09-11
feat(deps): TRAC-18-Upgrade-axios-version-in-node-tracer

eugene-lumigo:
- **feat(): upgrade axios version 1.7.7**

## Changes proposed
Had to modify `convertHeadersToKeyValuePairs` to convert `Record<string, string | ReadonlyArray<string>>` to `Record<string, string>`

```js
const headers: Record<string, string | ReadonlyArray<string>> = {
  "Set-Cookie": ["cookie1=value1", "cookie2=value2"]
};
```

## Note to reviewers
Reference to axios pull_request that cause breaking change to our code
https://github.com/axios/axios/pull/4322


src/hooks/fetch.ts
```diff
@@ -335,31 +335,43 @@ export class FetchInstrumentation {
...
    * Fetch library uses multiple format to represent headers, this function will convert them all to a key-value pair object.
-   * @param {[string, string][] | Record<string, string> | Headers} headers Headers object as used by the fetch library
+   * @param {string[][] | Record<string, string | ReadonlyArray<string>> | Headers} headers Headers object as used by the fetch library
    * @returns {Record<string, string>} The headers as a key-value pair object
    * @private
    */
   private static convertHeadersToKeyValuePairs(
     // @ts-ignore
-    headers: [string, string][] | Record<string, string> | Headers
+    headers: string[][] | Record<string, string | ReadonlyArray<string>> | Headers
```
harelmo-lumigo, line 344: Is the change from `[string, string][]` to `string[][]` also required from the Axios breaking change?
I think that  `[string, string][]` (an array of string pairs) is more precise than `string[][]` (an array of array of strings, that can technically also be `[['a'],['a'],['a']]` and not limited to pairs, so we better keep it if that's for us to decide

harelmo-lumigo (approved):
See my note about the type change - other than that, looks good 👍

src/hooks/fetch.ts
```diff
@@ -335,31 +335,43 @@ export class FetchInstrumentation {
...
-      const headersObject: Record<string, string> = {};
+    // Handle string[][] (array of key-value pairs)
+    else if (Array.isArray(headers)) {
       headers.forEach(([key, value]) => {
         headersObject[key] = value;
       });
-      return headersObject;
+    }
+    // Handle Record<string, string | ReadonlyArray<string>>
+    else if (typeof headers === 'object') {
```
sagivoululumigo, line 362: Find a better check to see if the `headers` object has entries (is it some sort of iteratable object for example)

sagivoululumigo (changes requested):
See comment

src/hooks/fetch.ts
eugene-lumigo: @harelmo-lumigo 
This is a driven by udici-type changed
from
`export type HeadersInit = [string, string][] | Record<string, string> | Headers`
to
`export type HeadersInit = string[][] | Record<string, string | ReadonlyArray<string>> | Headers`

src/hooks/fetch.ts
eugene-lumigo: @sagivoululumigo Done

src/hooks/fetch.ts
```diff
@@ -335,31 +335,43 @@ export class FetchInstrumentation {
...
-      const headersObject: Record<string, string> = {};
+    // Handle string[][] (array of key-value pairs)
+    else if (Array.isArray(headers)) {
       headers.forEach(([key, value]) => {
         headersObject[key] = value;
       });
-      return headersObject;
+    }
+    // Handle Record<string, string | ReadonlyArray<string>>
+    else if (headers && typeof headers === 'object' && !Array.isArray(headers)) {
```
sagivoululumigo, line 362: the `!Array.isArray(headers)` addition doesn't do anything, because it comes inside the else if block after the previous check, so we reach this check only if the headers var isn't an array

eugene-lumigo: Done

package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "@lumigo/node-core": "1.16.0",
     "agentkeepalive": "^4.1.4",
-    "axios": "1.6.5",
+    "axios": "1.7.7",
```
sagivoululumigo, line 29: You didn't update the lock file, make sure you update it using node 14 (If you use a newer version of node you will create a lock file that works for a newer version, we want to make it match the lowest version we support)

sagivoululumigo (changes requested):
Fix package lock file

sagivoululumigo (approved):
Approved

PR #517, opened 2024-10-03
fix: TRAC-33-crash-if-LUMIGO_SECRET_MASKING_REGEX-has-invalid-regex-exp

eugene-lumigo:


src/utils/payloadStringify.js
```diff
@@ -32,18 +32,27 @@ const keyToRegexes = (
...
+  const regexes =
+    tryParseEnvVar(backwardCompRegexEnvVarName) || tryParseEnvVar(regexesEnvVarName) || regexesList;
+
+  try {
+    return regexes.map((x) => new RegExp(x, 'i'));
+  } catch (e) {
+    regexes.forEach((x) =>
+      invalidMaskingRegexWarning(`Error processing regex: ${x} - ${e.message}`)
+    );
+    return fallbackRegexesList.map((x) => new RegExp(x, 'i'));
```
harelmo-lumigo, line 54: This will print a warning for all regexes, even for valid one, if a one or more in the list are invalid.
I suggest to use something like this:
```js
const parsedRegexes = regexes.map((x) => {
  try {
    return new RegExp(x, 'i')
  } catch (e) {
    invalidMaskingRegexWarning(`Error processing regex: ${x} - ${e.message}`)
    return e;
  }
});

if (parsedRegexes.some((x) => x instanceof Error)) {
  // print warning about using fallbacks
} else {
  return parsedRegexes
}
```
and print a message about using fallbacks, so it will be easier for the user to understand that the defaults patterns were used instead.

src/utils/payloadStringify.test.js
```diff
@@ -333,6 +333,17 @@ describe('payloadStringify', () => {
     ]);
   });
 
+  test('shallowMask -> requestBody -> LUMIGO_SECRET_MASKING_REGEX -> bad regex', () => {
+    utils.setDebug();
+    process.env[LUMIGO_SECRET_MASKING_REGEX] = '["a(a"]';
+    expect(shallowMask('requestBody', { a: 'b', aa: 'bla' })).toEqual({ a: 'b', aa: 'bla' });
+    expect(ConsoleWritesForTesting.getLogs()).toEqual([
+      expect.objectContaining({
+        msg: '#LUMIGO# - WARNING - "Failed to parse the given masking regex"',
```
harelmo-lumigo, line 342: This does not reflect the message you've added:
```js
invalidMaskingRegexWarning(`Error processing regex: ${x} - ${e.message}`)
```
are we testing the right thing here?

src/utils/payloadStringify.js
eugene-lumigo: I am about to use the same as in the `shallowMask` method

```
 try {
      givenSecretRegexes = JSON.parse(givenSecretRegexes);
      givenSecretRegexes = givenSecretRegexes.map((x) => new RegExp(x, 'i'));
    } catch (e) {
      invalidMaskingRegexWarning(e);
      givenSecretRegexes = null;
    }
```

I have it changed to

```
  try {
    return regexes.map((x) => new RegExp(x, 'i'));
  } catch (e) {
    invalidMaskingRegexWarning(e);
    logger.warn('Fallback to default regexes list', { fallbackRegexesList });
    return fallbackRegexesList.map((x) => new RegExp(x, 'i'));
  }
```

src/utils/payloadStringify.test.js
eugene-lumigo: updated, 
should be `invalidMaskingRegexWarning(e);`

src/utils/payloadStringify.js
harelmo-lumigo: Not sure I understand. What I meant is that we would like to print the regex value that was invalid, instead of letting the user look it up manually.
In the latter version it's still not printing the exact item in the list that's invalid, but just a generic message...

harelmo-lumigo (approved):

PR #518, opened 2024-10-08
feat: TRAC-40-adding-debug-logs-for-node-lambda-tracer-secret-masking

eugene-lumigo:
## Changes proposed

Adding debug logs for node lambda tracer secret masking

## Note to reviewers

[](https://lumigo.atlassian.net/browse/TRAC-40)


src/utils/payloadStringify.js
```diff
@@ -258,16 +268,34 @@ const invalidMaskingRegexWarning = runOneTimeWrapper((e) => {
 });
 
 const shallowMaskByRegex = (payload, regexes) => {
+  const byPassMaskingKeys = [
```
harelmo-lumigo, line 271: I would move this list to [this file](https://github.com/lumigo-io/lumigo-node/blob/13cf15fc85978bb8e6b2387c2fc4bb1d47c8ca7c/src/utils.ts#L23), and also use the constants of the env-var names instead of the names themselves - for consistency.
This will make a higher chance of us remembering to update that list once (and if) a new env-var is added

src/utils/payloadStringify.js
```diff
@@ -258,16 +268,34 @@ const invalidMaskingRegexWarning = runOneTimeWrapper((e) => {
...
     return payload;
   }
   if (typeof payload !== 'object') {
     logger.warn('Failed to mask payload, payload is not an object or string', payload);
     return payload;
   }
   return Object.keys(payload).reduce((acc, key) => {
-    if (keyContainsRegex(regexes, key)) {
+    if (byPassMaskingKeys.includes(key)) {
+      logger.debug('Bypass masking key', key);
```
harelmo-lumigo, line 295:
```suggestion
      logger.debug('Skipping masking of a Lumigo env-var', key);
```

src/utils/payloadStringify.js
```diff
@@ -277,6 +305,7 @@ const shallowMaskByRegex = (payload, regexes) => {
 };
 
 export const shallowMask = (context, payload) => {
+  logger.debug('Shallow masking payload', { context, payload });
```
harelmo-lumigo, line 308: Logging the entire payload might incur a high Cloudwatch cost. I think logging `payloadKeys: Object.keys(payload)` instead might be better without losing important data in this context

src/utils/payloadStringify.test.js
```diff
@@ -3,6 +3,7 @@ import {
   LUMIGO_SECRET_MASKING_ALL_MAGIC,
   LUMIGO_SECRET_MASKING_EXACT_PATH,
   LUMIGO_SECRET_MASKING_REGEX,
+  OMITTING_KEYS_REGEXES,
```
harelmo-lumigo, line 6: I think we don't need this imported here - any value would have been masked before the fix, and any value would be kept after it, so it's fine to some dummy const value - using this special value confused my a bit as I thought it has a special meaning in this case.

harelmo-lumigo (approved):

src/utils/payloadStringify.js
eugene-lumigo: Done

src/utils/payloadStringify.js
eugene-lumigo: Done

src/utils/payloadStringify.test.js
eugene-lumigo: Done

src/utils/payloadStringify.js
eugene-lumigo: Done

PR #519, opened 2024-10-15
fix: support new x ray format nodejs

sagivoululumigo:
## Fixes Issue

AWS is planning on releasing a change to the format of the X-Ray Trace ID value, and we need to support the new format (And the old one to support the transition phase).

AWS is removing the `Parent` and `Sampled` fields from any lambda that is set to passive tracing (the default).
The old format is:
```
Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1;Lineage=a87bd80c:1|68fd508a:5|c512fbe3:2
```

And the new format will be:
```
Root=1-5759e988-bd862e3fe1be46a994272793;Lineage=a87bd80c:1|68fd508a:5|c512fbe3:2
```

Notes:
* We use the `Root` value for the transaction id value (The last part splitting by `-`)
* We don't use the `Parent` field in the tracer code for anything except adding it to the lambda invocation http span `traceId` json. The python tracer doesn't add this field to the span, so I suspect it isn't used
* Parsing the x ray trace-id is crucial for the tracer to work, we had a bug with this parsing a year ago that caused the customers lambda to crash once the traceId format changed... need to be carful here

### DIFF
--- a/src/utils.test.js
+++ b/src/utils.test.js
@@ -181,6 +181,11 @@ describe('utils', () => {
     ${'Root=1-5b1d2450-6ac46730d346cad0e53f89d0'}                                                        | ${{ Root: '1-5b1d2450-6ac46730d346cad0e53f89d0' }}
     ${'Root=1-670d0060-1b85fdcd75ed1c2557382245;Lineage=1:aba0be3a:0'}                                   | ${{ Root: '1-670d0060-1b85fdcd75ed1c2557382245', Lineage: '1:aba0be3a:0' }}
     ${'Root=1-5b1d2450-6ac46730d346cad0e53f89d0;Parent=59fa1aeb03c2ec1f;Sampled=1;Lineage=1:aba0be3a:0'} | ${{ Root: '1-5b1d2450-6ac46730d346cad0e53f89d0', Parent: '59fa1aeb03c2ec1f', Sampled: '1', Lineage: '1:aba0be3a:0' }}
+    ${';;;;;;;;;'}                                                                                       | ${{}}
+    ${''}                                                                                                | ${{}}
+    ${'========='}                                                                                       | ${{}}
+    ${'=;=;=;=;'}                                                                                        | ${{}}
+    ${';;;;;===='}                                                                                       | ${{}}
   `('splitXrayTraceIdToFields', ({ xrayTraceId, expected }) => {
     expect(utils.splitXrayTraceIdToFields(xrayTraceId)).toEqual(expected);
   });
--- a/src/utils.test.js
+++ b/src/utils.test.js
@@ -162,7 +162,12 @@ describe('utils', () => {
     });
 
     // The x-ray value is too long, too many fields. We should parse it as usual
-const longXrayId = ['Root=1-5b1d2450-6ac46730d346cad0e53f89d0;Parent=59fa1aeb03c2ec1f;Sampled=1', ...Array(1000).fill(0).map((_, i) => `${i}=${i}`)].join(';')
+    const longXrayId = [
+      'Root=1-5b1d2450-6ac46730d346cad0e53f89d0;Parent=59fa1aeb03c2ec1f;Sampled=1',
+      ...Array(1000)
+        .fill(0)
+        .map((_, i) => `${i}=${i}`),
+    ].join(';');
     expect(utils.getTraceId(longXrayId)).toEqual({
       Parent: '59fa1aeb03c2ec1f',
       Root: '1-5b1d2450-6ac46730d346cad0e53f89d0',
@@ -188,10 +193,10 @@ const longXrayId = ['Root=1-5b1d2450-6ac46730d346cad0e53f89d0;Parent=59fa1aeb03c
   });
 
   test('splitXrayTraceIdToFields - too many fields', () => {
-    let longXrayId = '0=0';
-    for (let i = 1; i < 1000; i++) {
-      longXrayId += `;${i}=${i}`;
-    }
+    let longXrayId = Array(1000)
+      .fill(0)
+      .map((_, i) => `${i}=${i}`)
+      .join(';');
 
     const expectedFields = {};
     for (let i = 0; i < 100; i++) {
--- a/src/utils.test.js
+++ b/src/utils.test.js
@@ -84,62 +84,62 @@ describe('utils', () => {
   });
 
   test('getTraceId', () => {
-    // const awsXAmznTraceId =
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -97,7 +105,93 @@ export const getTracerInfo = (): { name: string; version: string } => {
   return { name, version };
 };
 
+const xRayFieldKeyValuePattern = /([^=;]+)=([^;]*)/g;
```
harelmo-lumigo, line 108: As we discussed, we can simplify this into:
```js
const regex = /([^;=]+)=([^;=]+)/g;
const matches = Array.from(str.matchAll(regex)).slice(0, 5); // support up to 5 keys?
const result = Object.fromEntries(matches.map(match => [match[1], match[2]]));
```

src/utils.ts
```diff
@@ -97,7 +105,93 @@ export const getTracerInfo = (): { name: string; version: string } => {
...
+export const splitXrayTraceIdToFields = (
+  awsXAmznTraceId: string,
+  maxFields: number = 100
+): { [key: string]: string } => {
+  const result: { [key: string]: string | null } = {};
+  let match: RegExpExecArray | null;
+  let iterations = 0;
+
+  // Resetting the regex lastIndex to ensure it works correctly for different inputs
+  xRayFieldKeyValuePattern.lastIndex = 0;
```
harelmo-lumigo, line 119: Might be irrelevant if you switch to the abbreviated method above, but anyway - why is this required?

src/utils.ts
```diff
@@ -180,10 +274,17 @@ export const getTraceId = (awsXAmznTraceId) => {
...
   const currentTime = Math.floor(Date.now() / 1000).toString(16);
-  return `Root=${rootArr[0]}-${currentTime}-${transactionId};Parent=${Parent};Sampled=${Sampled}`;
+  let patchedTraceId = `Root=${rootArr[0]}-${currentTime}-${transactionId};Parent=${Parent}`;
+  if (Sampled) {
+    patchedTraceId += `;Sampled=${Sampled}`;
+  }
+  if (Lineage) {
+    patchedTraceId += `;Lineage=${Lineage}`;
+  }
+  return patchedTraceId;
```
harelmo-lumigo, line 287: I think that concatenating here is a bit tricky with the leading `;`, WDYT of this:
```suggestion
  const patchedTraceId = [`Root=${rootArr[0]}-${currentTime}-${transactionId}`, `Parent=${Parent}`]
  if (Sampled) {
    patchedTraceId.push(`Sampled=${Sampled}`);
  }
  if (Lineage) {
    patchedTraceId.push(`Lineage=${Lineage}`);
  }
  return patchedTraceId.join(';');
```
?

src/utils.ts
```diff
@@ -97,7 +105,93 @@ export const getTracerInfo = (): { name: string; version: string } => {
...
+    throw new Error(
+      `X-Ray trace ID Root field is not in the expected format (Root=${fields.Root}, _X_AMZN_TRACE_ID=${awsXAmznTraceId})`
+    );
+  }
+  const transactionId = fields.Root.split('-')[2];
+  // Note: we might not need to generate a Parent field if it's not present,
+  // but for now we'll keep it as is to minimize changes. The python tracer doesn't generate it FYI.
+  const parent = fields.Parent || getRandomString(16);
+  const sampled = fields.Sampled;
+  const lineage = fields.Lineage;
```
harelmo-lumigo, line 165: There's a way to deconstruct ("unpack"), rename and provide a default value all at once, in case you want to reduce assignments:
```suggestion
  const { Parent: parent = getRandomString(16), Sampled: sampled,  Lineage: lineage } = fields;
```

src/utils.ts
```diff
@@ -97,7 +105,93 @@ export const getTracerInfo = (): { name: string; version: string } => {
...
+    Root: root,
+    transactionId,
+    Parent: parent,
+  };
+  if (sampled) {
+    parsedTraceId.Sampled = sampled;
+  }
+  if (lineage) {
+    parsedTraceId.Lineage = lineage;
+  }
```
harelmo-lumigo, line 177: Why not:
```suggestion
  const parsedTraceId: xRayTraceIdFields = {
    Root: root,
    transactionId,
    Parent: parent,
    Sampled: sampled,
    Lineage: lineage
  };
```
?
anyway the caller would need to check if the values are `undefined` when using the result, and it's way shorter

src/utils.test.js
```diff
@@ -132,6 +132,73 @@ describe('utils', () => {
...
+    }
+    expect(utils.getTraceId(longXrayId)).toEqual({
+      Parent: '59fa1aeb03c2ec1f',
+      Root: '1-5b1d2450-6ac46730d346cad0e53f89d0',
+      Sampled: '1',
+      transactionId: '6ac46730d346cad0e53f89d0',
+    });
+  });
+
+  test.each`
```
harelmo-lumigo, line 177: I would also test some crazy edge cases:
1. `;;;;;;;;;` ...
2. `=====` ...
3. `=;=;=;=;` ...
4. `;;;;;====` ...

harelmo-lumigo (changes requested):

src/utils.ts
sagivoululumigo: removed, switched to the new implementation

harelmo-lumigo (approved):

src/utils.ts
sagivoululumigo: Done

src/utils.ts
sagivoululumigo: I didn't want to add a `Sampled` or `Lineage` key with value `undefined`, I want to not add those keys if there is no value

src/utils.test.js
```diff
@@ -132,6 +132,73 @@ describe('utils', () => {
...
+    ${'Root=1-5b1d2450-6ac46730d346cad0e53f89d0;Parent=59fa1aeb03c2ec1f;Sampled=1;Lineage=1:aba0be3a:0'} | ${{ Root: '1-5b1d2450-6ac46730d346cad0e53f89d0', Parent: '59fa1aeb03c2ec1f', Sampled: '1', Lineage: '1:aba0be3a:0' }}
+  `('splitXrayTraceIdToFields', ({ xrayTraceId, expected }) => {
+    expect(utils.splitXrayTraceIdToFields(xrayTraceId)).toEqual(expected);
+  });
+
+  test('splitXrayTraceIdToFields - too many fields', () => {
+    let longXrayId = '0=0';
+    for (let i = 1; i < 1000; i++) {
+      longXrayId += `;${i}=${i}`;
+    }
```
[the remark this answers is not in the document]
sagivoululumigo: Done

PR #522, opened 2024-10-30
fix: RD-13884-Node-Tracer-can-change-event-body-of-a-Lambda-function

eugene-lumigo:


src/parsers/eventParser.ts
```diff
@@ -254,7 +254,7 @@ export const parseEvent = (event) => {
   if (isCloudfrontEvent(event)) {
     return parseCloudfrontEvent(event);
   }
-  return event;
+  return event ? { ...event } : null;
```
harelmo-lumigo, line 257: Note that `...` is a shallow copy, so if we somehow scrub deep levels, the payload will still be overridden.
To mitigate this, [cloneDeep](https://www.geeksforgeeks.org/lodash-_-clonedeep-method/) can be used - but has some performance impact when dealing with large payloads.
If we decide that shallow copy is fine, let's document this here for future issues.

harelmo-lumigo (approved):

src/parsers/eventParser.ts
eugene-lumigo: Done

PR #523, opened 2024-11-06
feat: RD-13971-Masking-does-not-work-for-outbound-JSON-requests

eugene-lumigo:
- **fix: logging improvments,**
- **fix: scrub should work for JSON payloads without content-type**


src/reporter.ts
```diff
@@ -76,7 +76,19 @@ export const sendSpans = async (spans: any[], addEnrichmentSpan: boolean = true)
...
+      JSON.parse(str);
+      return true;
+    } catch (e) {
+      return false;
+    }
+  };
+
+  return (
+    (isString(payload) && headers['content-type'] && headers['content-type'].includes('json')) ||
+    isJsonString(payload)
```
harelmo-lumigo, line 90: I suggest several improvements:
1. shorter conditional on the header
2. lowercase check, for cases it's not normalized
3. moving the string check to be first, so if it fails there no need to call `isJsonString` for no reason
```suggestion
    isString(payload) && (headers['content-type']?.toLowerCase().includes('json') || isJsonString(payload)
```

harelmo-lumigo (approved):

src/reporter.ts
```diff
@@ -75,8 +75,29 @@ export const sendSpans = async (spans: any[], addEnrichmentSpan: boolean = true)
   safeExecute(logSpans)(rtt, spans);
 };
 
+const isNotJsonString = (str: string) => {
```
harelmo-lumigo, line 78: A few notes on that:
1. I think that having it named `isNot` and negating the result is confusing... IMO it's better to name it `isJsonObjectOrArray`, remove the negation so it would be:
```
  return (
    (trimmed.startsWith('{') && trimmed.endsWith('}')) ||
    (trimmed.startsWith('[') && trimmed.endsWith(']'))
  );
``` 
and negate the result on the caller side.
Also, I've suggested to rename it to `isJsonObjectOrArray` since `1` or `true`, for instance, are also valid JSON values but won't be identified as such by the `{,},[,]` check. Naming it that way will express that it's focused in objects and arrays only

eugene-lumigo: 1. Agree on that. With this function, I am trying to reduce the number of JSON.parse() operation

PR #528, opened 2024-11-30
fix: RD-14191-SECRET_MASKING_EXACT_PATH-change-the-response-body-of-the-function

eugene-lumigo:


src/spans/awsSpan.ts
```diff
@@ -51,6 +51,8 @@ export const NEO4J_SPAN = 'neo4j';
 export const ENRICHMENT_SPAN = 'enrichment';
 export const PRISMA_SPAN = 'prisma';
 
+const clone = require('rfdc')({ proto: true });
```
harelmo-lumigo, line 54: What does the `proto: true` mean?

src/spans/awsSpan.ts
```diff
@@ -242,7 +244,7 @@ export const getEndFunctionSpan = (functionSpan: GenericSpan, handlerReturnValue
   const ended = new Date().getTime();
   let returnValue: any;
   try {
-    returnValue = payloadStringify(data);
+    returnValue = payloadStringify(clone(data));
```
harelmo-lumigo, line 247: Maybe we should the `clone` inside `payloadStringify`? if we're concerned about it recursively cloning every subset of the object, let's make it:
```ts
const payloadStringify = payload => originalPayloadStringify(clone(data))
```
that way we won't have potential issue in otherexisting of future use cases of `payloadStringify`. WDYT?

src/spans/awsSpan.test.js
```diff
@@ -444,6 +445,35 @@ describe('awsSpan', () => {
...
+        string: 'value',
+        object: {
+          string: 'value',
+          object: {
+            string: 'value',
+          },
+        },
+      },
+    };
+    const expectedData = clone(handlerReturnValue.data);
```
harelmo-lumigo, line 465: Why do we need the `clone` here? can't we just check that fields we focus on are not changed to `****`?

src/spans/awsSpan.ts
eugene-lumigo: I'll revert it

harelmo-lumigo: why revert? sounds like a good thing :)

src/spans/awsSpan.ts
eugene-lumigo: @harelmo-lumigo does make sense. However, in this approach, we will need to calculate the time impact. 
I propose to have it inside `payloadStringify` but only in the 
```
if (getSecretMaskingExactPath()) {}
```
section. As we know it is implementation `LUMIGO_SECRET_MASKING_EXACT_PATH`

harelmo-lumigo: For now, let's not change it and keep it that way.
I'd suggest to add a ticket for considering the clone at a single point in the process later, as a separate change we can benchmark easily

harelmo-lumigo (approved):

PR #533, opened 2025-01-12
feat: RD-14290-support-AWS-Lambda-Stream

eugene-lumigo:


src/tracer/tracer.test.js
```diff
@@ -818,18 +818,18 @@ describe('tracer', () => {
...
 
     const decoratedUserHandler = tracer.trace({})(handler);
-    await decoratedUserHandler(event, context);
+    const result = await decoratedUserHandler(event, responseStream, context);
 
     expect(decoratedUserHandler[HANDLER_STREAMING]).toEqual(STREAM_RESPONSE);
-    expect(spies.warnClient).toHaveBeenCalledWith(
-      'Tracer is disabled, running on a response stream function'
-    );
+    expect(result).not.toBeNaN();
```
harelmo-lumigo, line 832: What's the thing with the `NaN`? not sure I got it

testUtils/handlerInputsBuilder.ts
```diff
@@ -45,8 +46,14 @@ export class HandlerInputsBuilder {
     return this;
   };
 
+  withResponseStream = ()=>{
+    this._responseStream = {}
```
harelmo-lumigo, line 50: Aren't we supposed to use a real stream here? not sure I understand

src/tracer/tracer.test.js
eugene-lumigo: changed to

`expect(result).toEqual({ hello: 'world' });`

testUtils/handlerInputsBuilder.ts
eugene-lumigo: we don't actually need ResponseStream class. I am removing it, as it is confusing. 
I have test updated to to be able to inject stream, but in our case we just want to test proper handler is used and responseStream is used.

eugene-lumigo:
@harelmo-lumigo please approve

harelmo-lumigo (approved):

